PR #8, opened 2021-11-16
Encapsulate validation

charles-cowart:
Work-in-progress.
Supersedes PR#7.
Sample-sheet validation will now be handled by a new method in mg-script's Pipeline class.
Pipeline class will perform any additional validation not presently handled by metapool.
metapool is being removed as a dependency of qp-klp.


charles-cowart:
Ready for review.

PR #12, opened 2021-12-16
Fixes based on pre-production tests

charles-cowart:
checking return-codes for mv, cp, and tar operations removed due to a number of 'false-positives' encountered.
removed double-conversion of list to string in situations where sif files were being generated.

### DIFF
--- a/qp_klp/klp.py
+++ b/qp_klp/klp.py
@@ -219,14 +219,6 @@ def sequence_processing_pipeline(qclient, job_id, parameters, out_dir):
         qclient.update_job_step(job_id, "Step 6 of 6: Copying results to "
                                         "archive")
 
-        # just use the filenames for tarballing the sifs.
-        # this will prevent the tarball from having an arbitrarily nested
-        # tree.
-        # the sifs should all be stored in the {out_dir} by default.
-        sifs = [basename(x) for x in sifs]
-        # convert sifs into a list of filenames.
-        sifs = ' '.join(sifs)
-
         cmds = [f'cd {out_dir}; tar zcvf logs-ConvertJob.tgz ConvertJob/logs',
                 f'cd {out_dir}; tar zcvf reports-ConvertJob.tgz '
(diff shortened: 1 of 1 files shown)


qp_klp/klp.py
```diff
@@ -282,13 +272,15 @@ def sequence_processing_pipeline(qclient, job_id, parameters, out_dir):
...
+        with open(cmd_log_fp, 'w') as cmd_log_f:
+            for cmd in cmds:
+                cmd_log_f.write(f'{cmd}\n')
+
         for cmd in cmds:
             p = Popen(cmd, universal_newlines=True, shell=True,
                       stdout=PIPE, stderr=PIPE)
             std_out, std_err = p.communicate()
-            return_code = p.returncode
-            if return_code != 0:
```
antgonza, line 290: Just to confirm, why isn't this required anymore? AFAIK this is good to have ...

charles-cowart: I would say in general it's good, but in practice it depends on how well-behaved the command being called is. This goes back to the recent issue where tar returned 1 ('cd /redacted/redacted/working_dir/d6624327-bd4e-4702-9bb2-06597dce7892; tar zcvf logs-FastQCJob.tgz FastQCJob/logs') even though logs-FastQCJob.tgz was properly generated and the unarchived files were indistinguishable from a manually generated tgz file that properly returned 0.

For an entire run to error out after successfully iterating through all of the stages seems misaligned. At best the system should return a warning, rather than an error.

antgonza: I'll say, let's readd it ...

charles-cowart: and return a warning?

antgonza: Nah, I'll say break the full job cause this was actually a bug in the storage, which was fixed this morning.

charles-cowart: okay.

PR #13, opened 2021-12-17
Move sif tarball to final_results and files to upload_dir

charles-cowart:


qp_klp/klp.py
```diff
@@ -246,6 +246,10 @@ def sequence_processing_pipeline(qclient, job_id, parameters, out_dir):
                 csv_fps.append(join(root, csv_file))
 
         for project, upload_dir in special_map:
+            if sifs:
+                # move uncompressed sifs to upload_dir.
+                cmds.append(f'cd {out_dir}; mv {tmp} {upload_dir}')
```
antgonza, line 251: Note that this will move _all_ files to the first upload_dir, and it should be one per project.

charles-cowart: 🤦

qp_klp/tests/test_klp.py
```diff
@@ -341,6 +344,18 @@ def test_sequence_processing_pipeline(self):
 
         self.assertEqual(ainfo, exp)
 
+        # verify cmd.log
```
antgonza, line 347: I think it will be better to have a multiline string with all the commands that are expected so it's easier to read while reviewing the code; what do you think?

antgonza:
@charles-cowart, just a friendly ping about improving the test.

qp_klp/klp.py
```diff
@@ -246,6 +246,12 @@ def sequence_processing_pipeline(qclient, job_id, parameters, out_dir):
                 csv_fps.append(join(root, csv_file))
 
         for project, upload_dir in special_map:
+            if sifs:
```
antgonza, line 249: Note that this will fail if a project has no blanks but other do; you need to check per specific project.

antgonza: Not pretty but this works:
``` if sifs and [sif for sif in sifs if f'{project}_blanks.tsv' in sif]```

qp_klp/tests/test_klp.py
charles-cowart: I agree! I think that's a good idea. I just added that to the test.

antgonza (approved):

PR #18, opened 2022-03-01
fix #14

### LINKED ISSUE #14: Add lane number via GUI
The GUI should allow to select a number that represents the lane number and should be propagated to the sample_sheet; this should allow users to not have to add it manually.

antgonza:


antgonza:
@charles-cowart, this is ready for review; thank you.

charles-cowart (approved):
Looks good. No changes needed.

qp_klp/klp.py
```diff
@@ -79,7 +80,20 @@ def sequence_processing_pipeline(qclient, job_id, parameters, out_dir):
...
+                    add_lane_number = False
+
+                if add_lane_number:
+                    d = d.split(',')
+                    d[0] = str(lane_number)
+                    d = ','.join(d)
+                f.write(f'{d}\n')
+
+                if d.startswith('Lane,'):
+                    add_lane_number = True
```
ElDeveloper, line 96: Using metapool you could have also done:

```
from metapool import KLSampleSheet

# this assumes the csv file is readable at that path
ss = KLSampleSheet(sample_sheet_path)
for sample in ss:
    sample['Lane'] = '8675309'
ss.write('test.csv')
```

antgonza: Good to know; thank you. A couple of questions:
1. do you think is worth changing?
2. does KLSampleSheet also work with strings? If no, does this mean that I'll need to write the contents of `sample_sheet['body']` in a `temp` file and then read it again?

ElDeveloper: 1. Probably not worth changing now if this is working - if it ain't broken don't fix it 🤷‍♂️. My suggestion is mostly to bring awarenes to the fact that `KLSampleSheet` can handle operations on the data in more straight-forward ways. ✨ 
2. It currently doesn't, as you say you would need to write the file to a temporary file.  That being said, looking at the code it wouldn't be too bad to change the parse method to get it to work with a `StringIO` object (so we could avoid writing -> reading).

PR #22, opened 2022-03-11
Updated to reflect Pipeline changes + Failed Samples

charles-cowart:
Pipeline now requires a sample-sheet as a parameter during object creation. Some changes were made to this klp.py to reflect this.
Job objects are now able to audit their results based on a provided list of sample ids and return a list of the missing ones.
klp.py will now write out the list of samples that failed to process, and at what stage did they fail processing.


antgonza (approved):
Thank you.

PR #30, opened 2022-04-06
Convert sample_names to tube_ids in prep-file output

charles-cowart:
If a study in Qiita has tube-ids, generate a map of those tube-ids to sample_names (qiita version, w/out leading zeroes) and pass that data to GenPrepFileJob.
GenPrepFileJob will take the optional parameter and perform the conversion on the fly before tarballing the prep-files as normal.
This also makes it easier to test this functionality, since unit-tests are not connecting to Qiita.


charles-cowart:
I am going to write this in a function as part of the qp-klp project instead.

PR #31, opened 2022-04-06
Map sample names to tube ids

charles-cowart:
qp-klp can now scan Qiita for tube-ids and if they are present with a study, it can modify prep-files on-the-fly to reflect the proper sample-name.

### DIFF
--- a/qp_klp/klp.py
+++ b/qp_klp/klp.py
@@ -145,6 +145,7 @@ def sequence_processing_pipeline(qclient, job_id, parameters, out_dir):
 
         sheet_df = sample_sheet_to_dataframe(sheet)
         errors = []
+        sn_tid_map_by_project = {}
         for project, _df in sheet_df.groupby('sample_project'):
             project_name = remove_qiita_id(project)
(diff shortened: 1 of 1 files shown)


qp_klp/klp.py
```diff
@@ -154,6 +186,7 @@ def _update_current_message(msg):
             qsamples = {
                 s.replace(f'{qiita_id}.', '') for s in qclient.get(qurl)}
             sample_name_diff = sheet_samples - qsamples
+            sn_tid_map_by_project[project_name] = None
```
antgonza, line 189: I think you are missing a `if sn_tid_map_by_project[project_name] is not None` somewhere in the code ... no?

charles-cowart: The FailedSamplesRecord is separate from the prep files that seqpro generates. map_sample_names_to_tube_ids() is just there to modify the prep files after seqpro creates them. If we need to modify the sample names for failed_samples.html and the other html file, we can do that as a next step.

antgonza (approved):
Thank you @charles-cowart !

PR #34, opened 2022-10-04
bugfix

charles-cowart:
Fix error reporting in Qiita interface when there are less than five error messages.
Hotfixed this into production after Shrikant encountered this error.

### DIFF
--- a/qp_klp/klp.py
+++ b/qp_klp/klp.py
@@ -226,14 +226,29 @@ def sequence_processing_pipeline(qclient, job_id, parameters, out_dir):
                     if not tube_id_diff:
                         continue
                     len_tube_id_overlap = len(tube_id_diff)
-                    tids_example = ', '.join(choices(list(tids), k=5))
+
+                    tids = list(tids)
+
+                    if len(tids) > 5:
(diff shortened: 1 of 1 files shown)


antgonza:
@charles-cowart, this can be closed, right?

charles-cowart:
@antgonza It can be closed, but a new one should be issued. This code has migrated from klp.py to process_amplicon and process_metagenomics and the main branch does not show the change.

antgonza:
I thought these were those lines: https://github.com/qiita-spots/qp-knight-lab-processing/blob/main/qp_klp/process_amplicon_job.py#L106

charles-cowart:
> I thought these were those lines: https://github.com/qiita-spots/qp-knight-lab-processing/blob/main/qp_klp/process_amplicon_job.py#L106

No with that line as it is, choices will fail if tids is less than five elements. Moreover, I just checked and confirmed choices() can select the same element more than once. I'm going to issue a new PR that uses random.sample() instead:

result = tids if len(tids) < 6 else sample(tids, k=5)

charles-cowart:
Closing. Superseded by https://github.com/qiita-spots/qp-knight-lab-processing/pull/46

PR #37, opened 2023-02-15
Refactor qp-klp

charles-cowart:
Refactor qp-klp to accept amplicon-pipeline in the future. Clean up and capture all Errors so they can correctly be passed back as success = False and an error-message (No compressed stack-frame msgs).

Original PR has been broken up into two parts to make it easier to conceptualize and review.

### DIFF
--- a/qp_klp/klp.py
+++ b/qp_klp/klp.py
@@ -15,11 +15,9 @@ from qp_klp.process_metagenomics_job import process_metagenomics
 from sequence_processing_pipeline.Pipeline import Pipeline
 from sequence_processing_pipeline.PipelineError import PipelineError
 from qp_klp.klp_util import StatusUpdate
-from traceback import print_exc
 
 
 CONFIG_FP = environ["QP_KLP_CONFIG_FP"]
-LOG_FP = environ["QP_KLP_LOG_FP"]
 
 
 def sequence_processing_pipeline(qclient, job_id, parameters, out_dir):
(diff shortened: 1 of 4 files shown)


qp_klp/tests/test_klp.py
```diff
@@ -319,7 +319,23 @@ def test_sequence_processing_pipeline(self):
         self.assertFalse(success)
 
         self.assertEqual(msg, "This doesn't appear to be a valid sample sheet"
-                              "; please review.")
+                              " or mapping file; please review.")
+
+    def dontest_sequence_processing_pipeline_success(self):
```
antgonza, line 324: is this intentional?

charles-cowart: My apologies, yes. Since qp-klp has mg-scripts as a dependency, it looks like some of the tests now fail because mg-scripts is more robust. For example, these tests create a directory for QCJob or FastQCJob output, but doesn't create fake fastq files. mg-scripts now detects that and returns an error message instead of treating it as a degenerate case. The tests here need to be reworked. Since Qiita and all of its dependencies are dependencies of the plugin, it was best to determine which tests failed here.

charles-cowart: It seems like the last time CI was run successfully was about ten months ago. I can look into a trigger to perform CI whenever mg-scripts or metapool changes.

charles-cowart:
@antgonza ready for review!

qp_klp/klp.py
```diff
@@ -136,391 +50,43 @@ def sequence_processing_pipeline(qclient, job_id, parameters, out_dir):
...
-            elif str(e).startswith("Sample-sheet has the following errors:"):
-                msg = str(e)
-                _update_current_message(msg)
-                raise ValueError(msg)
+    # StatusUpdate uses qclient to easily update the status-line and maintain
+    # a certain amount of state needed to provide the updates users are
+    # accustomed to. E.g. "Step 1 of 6: Setting up pipeline"
+    status_line = StatusUpdate(qclient, job_id)
+
+    try:
```
antgonza, line 58: This is an extremely long `try`, is it necessary? If yes, can you split in smaller blocks of try/except with specific exceptions in those smaller blocks? Using this kind of long try blocks hides what's actually raising the exceptions.

antgonza (changes requested):
Looks good, this is mainly code shifting to allow the new Amplicon code to be added. Only one comment.

qp_klp/klp.py
charles-cowart: Will do!

PR #38, opened 2023-02-15
basic amplicon pipeline added

charles-cowart:
Basic amplicon pipeline added.
Does not include sample_id validation.
Does not include prep-info-file generation.
Does not include tests.


charles-cowart:
Sample ID validation + Amplicon-specific unittests added. Ready for review.

qp_klp/process_amplicon_job.py
```diff
@@ -0,0 +1,391 @@
...
+                           config['job_max_array_length'],
+                           True)
+
+    if not skip_exec:
+        fastqc_job.run(callback=status_line.update_job_step)
+
+    status_line.update_current_message("Skipping Prep Info File Generation")
+
+    '''
+    Prep Information File generation disabled.
```
antgonza, line 243: 😭

qp_klp/process_amplicon_job.py
```diff
@@ -0,0 +1,391 @@
...
+from sequence_processing_pipeline.FastQCJob import FastQCJob
+# from sequence_processing_pipeline.GenPrepFileJob import GenPrepFileJob
+from sequence_processing_pipeline.Pipeline import Pipeline
+from sequence_processing_pipeline.PipelineError import PipelineError
+from subprocess import Popen, PIPE
+import pandas as pd
+import shutil
+
+
+def process_amplicon(mapping_file_path, qclient, run_identifier, out_dir,
```
antgonza, line 19: I'm actually surprised that the metagenomic and amplicon sections are so different that deserve their own full method - I would have thought that they were basically the same with a few if/else's, is that right?

antgonza (changes requested):
Some comments; thank you @charles-cowart !

qp_klp/process_amplicon_job.py
```diff
@@ -0,0 +1,391 @@
+from metapool.prep import remove_qiita_id
+from os import listdir, makedirs
+# from os import walk
```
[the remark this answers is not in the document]
charles-cowart: Will do. They are the dependencies for prep-info-file generation, so they'll come back eventually.

qp_klp/process_amplicon_job.py
charles-cowart: One careful step at a time. =)

qp_klp/process_amplicon_job.py
charles-cowart: I think that from a working amplicon pipeline we could mechanically merge them back together. Developing the amplicon pipeline while preserving the robustness of the metagenomic pipeline got a little hairy so I elected to separate the two. I also tried to change the Job objects as little as possible since they would either need to be shared or duplicated as well. That seemed to work out in that ConvertJob required no change and QCJob could be dropped entirely. FastQCJob required just a small change to support amplicon. Then it became a matter of tweaking the inputs and outputs between each Job so that's where the differences come in I suppose. Small changes like copying files into the directory structure FastQCJob was expecting were needed and not being able to generate failed-samples records became a thing.

charles-cowart: Mapping file functionality is currently a part of Pipeline and if the mapping file is really the analog of a sample-sheet, then the functionality would be better off in the metapool package.  The copying of files into QCJob's output file/directory structure could also be considered a degenerate case of QCJob operation so that could be refactored into QCJob as well. Those two items and possibly refactoring both lines into one or standardizing for potentially new pipelines in the future would be what comes to mind.

qp_klp/process_amplicon_job.py
```diff
@@ -0,0 +1,383 @@
...
+                tube_id_diff = mf_samples - tids
+                if not tube_id_diff:
+                    continue
+                len_tube_id_overlap = len(tube_id_diff)
+                tids_example = ', '.join(choices(list(tids), k=5))
+                error_tube_id = (
+                    f'tube_id in Qiita but {len_tube_id_overlap} missing '
+                    f'samples. Some samples from tube_id: {tids_example}.')
+            else:
+                error_tube_id = 'No tube_id column in Qiita.'
```
antgonza, line 105: I don't fully follow why we need these two blocks? Can they be merged into one?

antgonza (changes requested):
Some questions.

qp_klp/process_amplicon_job.py
charles-cowart: Good catch, ty!

qp_klp/process_amplicon_job.py
```diff
@@ -0,0 +1,383 @@
...
+    errors = []
+    sn_tid_map_by_project = {}
+
+    for project in projects:
+        # remove the qiita-id prepending the project_name
+        project_name = remove_qiita_id(project['project_name'])
+        qiita_id = project['qiita_id']
+
+        df = pipeline.mapping_file
+        df = df.loc[df['project_name'] == project_name]
```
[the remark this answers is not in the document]
charles-cowart: It's part of a filter to generate a set of sample-names that are part of the current project specified in the loop and aren't BLANKs.

PR #40, opened 2023-02-23
Verify study id exists

charles-cowart:
Notify the user when values in the project_name column of a user-supplied mapping-file are not appended with a Qiita ID.


antgonza:
Can you add a tests for the new code?

charles-cowart:
@antgonza ready for review!

qp_klp/process_amplicon_job.py
```diff
@@ -46,14 +46,25 @@ def process_amplicon(mapping_file_path, qclient, run_identifier, out_dir,
...
+                                "be appended with a Qiita ID.")
+
         # assume the BLANKS in the mapping-file are not prepended w/qiita-id
         # or some other value. Confirmed w/wet-lab.
         df = pipeline.mapping_file
-        df = df[df['project_name'] == project_name]
+        df = df[df['project_name'] == project_name_with_qid]
+
+        if df.shape[0] == 0:
+            raise PipelineError(f"project '{project_name_with_qid}' has zero "
```
antgonza, line 65: This error doesn't have a tests, can you add one? BTW do we really need the shape in the error?

antgonza (changes requested):

qp_klp/process_amplicon_job.py
charles-cowart: @antgonza After reviewing it, I'm actually going to remove it. That situation was able to happen because `df = df[df['project_name'] == project_name_with_qid]` was instead `df = df[df['project_name'] == project_name]`. It was possible for a mapping-file to contain project_names w/qiita-ids appended and get an empty df with the above statement because we're trying to match on just the project_name w/out qiita_id.

Now that it's been corrected, unlike w/sample-sheets, it's not possible to declare a project_name that won't have a sample in it; it doesn't have a separate [data] and [Bioinformatics] sections.

charles-cowart: Specifically, the list of project-names itself comes from scanning the same column so I can't define a project without defining a sample.

antgonza (approved):

PR #41, opened 2023-02-24
Switch added to use bcl2fastq instead of bcl-convert

charles-cowart:


charles-cowart:
Right now the switch is hardcoded in the code. Can see it going into configuration or as a constructor parameter.

antgonza (approved):

PR #45, opened 2023-03-13
WIP Updated to use GenPrepFileJob._get_prep_file_paths metadata

charles-cowart:


qp_klp/tests/test_klp.py
```diff
@@ -617,29 +619,30 @@ def test_map_sample_names_to_tube_ids(self):
...
+        for prep_file in files_list:
+            df = pd.read_csv(prep_file, delimiter='\t')
+            # confirm .tsv file loaded everything as strings
+            types = dict(df.dtypes)
+            unique_types = set([types[k] for k in types])
+            bad_types_present = False
+            for type in unique_types:
+                if type != object:
+                    bad_types_present = True
+                    break
```
antgonza, line 636: In read_csv you can add dtype=str and you could avoid all these, no?

qp_klp/tests/test_klp.py
```diff
@@ -617,29 +619,30 @@ def test_map_sample_names_to_tube_ids(self):
...
+            res = df['sample_name'].tolist()
+            res = [x for x in res if not x.startswith('tube_id')]
 
-                self.assertEquals(len(res), 0)
+            self.assertEquals(len(res), 0)
 
-                # write modified results back out to file
-                df.to_csv(prep_file, index=False, sep="\t")
+            # write modified results back out to file
+            df.to_csv(prep_file, index=False, sep="\t")
```
antgonza, line 645: Is there a teste here? I think this only converts files, no?

antgonza (changes requested):

qp_klp/tests/test_klp.py
charles-cowart: I think the intention then was to somehow confirm all columns were loaded correctly. I actually think it should be removed because it doesn't actually test anything that can't be forced by dtype=str as you mentioned. It's actually testing pandas more than the code.

qp_klp/tests/test_klp.py
charles-cowart: I updated map_sample_names_to_tube_ids() and the test for it. The latter seems much more sensible now.

qp_klp/klp_util.py
```diff
@@ -1,44 +1,30 @@
...
-                            sample_name = sn_tid_map_by_proj[project][
-                                sample_name]
-                            df.at[i, "sample_name"] = sample_name
-                    results[project][prep_file] = df
+                            smpl_name = smpl_name.lstrip('0')
+                            if smpl_name in sn_tid_map_by_proj[proj]:
+                                tube_id = sn_tid_map_by_proj[proj][smpl_name]
+                                df.at[i, "sample_name"] = tube_id
+                    df.to_csv(prep_file, index=False, sep="\t")
                     break
```
antgonza, line 26: Just to confirm, why do we need this break?

qp_klp/process_metagenomics_job.py
```diff
@@ -245,17 +246,22 @@ def process_metagenomics(sample_sheet_path, lane_number, qclient,
...
+        # concatenate the lists of paths across all study_ids into a single
+        # list.
+        pfp_list = list(chain.from_iterable(prep_file_paths.values()))
+    else:
+        # under testing conditions, gpf_job.prep_file_paths will not be
+        # populated. Generate pfp_list from walking the expected location.
+        pfp_list = []
+        for root, dirs, files in walk(join(pipeline.output_path,
+                                           'GenPrepFileJob', 'PrepFiles')):
+            for prep_file in files:
```
antgonza, line 260: This looks like it's looking for files, is that right? I thought the plan was to not do that, no?

antgonza (commented):
Some questions ...

qp_klp/klp_util.py
charles-cowart: we don't. :)

qp_klp/process_metagenomics_job.py
charles-cowart: The issue is that skip_exec is True during testing and GenPrepFileJob.run() and thus seqpro will not execute. If seqpro doesn't execute there won't be a list of known prep-file outputs. Hence, searching for prep-files in the designated location is the only way to have a generalized solution for the pipeline.

PR #47, opened 2023-03-21
WIP Update SIFs w/additional sample_names

charles-cowart:


charles-cowart:
Working on a proper test.

charles-cowart:
@antgonza I changed the name of mg-scripts' Pipeline.generate_sample_information_files() to generate_sample_info_files() to clean up the code. this PR will pass CI again once the mg-scripts PR is merged.

qp_klp/process_amplicon_job.py
```diff
@@ -290,6 +291,8 @@ def process_amplicon(mapping_file_path, qclient, run_identifier, out_dir,
                         data_type = key
                         break
 
+                # add_sif_info.append(metadata['sample_name', 'project_name'])
```
antgonza, line 294: needed?

qp_klp/process_amplicon_job.py
```diff
@@ -298,6 +301,36 @@ def process_amplicon(mapping_file_path, qclient, run_identifier, out_dir,
                 prep_id = reply['prep']
 
                 touched_studies_prep_info[study_id].append(prep_id)
+
+        # convert additional_sif_info from a list of dataframes into a single
+        # merged frame before passing to generate_sample_information_files().
+        # add_sif_info = pd.concat([add_sif_info],
+        #                         ignore_index=True).drop_duplicates()
```
antgonza, line 308: needed?

antgonza (changes requested):
Some minor comments.

qp_klp/process_amplicon_job.py
charles-cowart: Nope. Originally I didn't think I needed to query Qiita for a list of BLANKs, so I had an alternate method. I kept it in as a point of discussion. I'll remove it.

qp_klp/process_amplicon_job.py
charles-cowart: same.

qp_klp/klp_util.py
```diff
@@ -1,5 +1,35 @@
...
+
+        # Get list of BLANKs already registered in Qiita.
+        from_qiita = qclient.get(f'/api/v1/study/{study_id}/samples')
+        from_qiita = [x for x in from_qiita if
+                      x.startswith(f'{study_id}.BLANK')]
+
+        # Generate list of BLANKs that need to be ADDED to Qiita.
+        new_blanks = (set(blanks) | set(from_qiita)) - set(from_qiita)
+
+        if len(new_blanks):
```
antgonza, line 25: Maybe I'm wrong but I think this will add dummy data to all columns of the blanks vs. _only_ dummy data for those columns that do _not_ exist in the local blanks; right?

charles-cowart: If I understand your question correctly, no I don't think it will add dummy data over and above what is already there. I've confirmed new_blanks will just be the subset of BLANKs found in the SIFs but not found in Qiita. Those BLANKs will be inserted w/dummy data, but the categories are all pre-existing so no new columns are being added to the study. The existing BLANKs in Qiita should already have (possibly dummy) data associated with them and they should not be overwritten by this PATCH, since it's not a POST.

charles-cowart:
@antgonza Ready for review! Information from the SIF files now become new metadata columns in the study.

antgonza (approved):

PR #57, opened 2023-04-21
Fix read csv in update_blanks_in_qiita

wasade:
The order of commits is inconsequentially backwards -- I didn't catch the unusual entry in `.gitignore` until a few commits in. 

This PR does the following:

- asserts we can provide numeric IDs to `update_blanks_in_qiita`. This is an edge case that we do not expect, but all the more reason to be defensive
- limits the `update_blanks_in_qiita` method to only consider samples with `blank` in their names

I don't have a full dev environment locally so unfortunately I do need to rely on CI for test suite execution.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -130,6 +130,3 @@ dmypy.json
 
 # Plugin configuration file
 configuration.json
-
-# test generated files
-qp_klp/tests/*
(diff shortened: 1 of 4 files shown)


.gitignore
```diff
@@ -130,6 +130,3 @@ dmypy.json
 
 # Plugin configuration file
 configuration.json
-
-# test generated files
-qp_klp/tests/*
```
charles-cowart, line 135: Why is this line being removed again?

charles-cowart: I believe it's to prevent example output files generated by the notebooks from being accidentally added to the repo. No worries on my end though. I can approve it as is.

charles-cowart (approved):

charles-cowart (approved):

PR #60, opened 2023-05-12
DRY refactoring

charles-cowart:
Goal is to refactor code so that there is minimal duplication, easily extensible to enable metatranscriptomics and future Assay types, and breakout all of the functionality from the main plugin function so they can be tested properly.


.gitignore
```diff
@@ -127,6 +127,3 @@ dmypy.json
 
 # Pyre type checker
 .pyre/
-
-# Plugin configuration file
-configuration.json
```
antgonza, line 132: why delete this?

qp_klp/AmpliconStep.py
```diff
@@ -0,0 +1,102 @@
...
+class AmpliconStep(Step):
+    def __init__(self, pipeline, master_qiita_job_id, sn_tid_map_by_project,
+                 status_update_callback=None):
+        super().__init__(pipeline,
+                         master_qiita_job_id,
+                         sn_tid_map_by_project,
+                         status_update_callback)
+
+        if pipeline.pipeline_type != 'amplicon':
+            raise ValueError("Cannot instantiate AmpliconStep object from "
```
antgonza, line 16: Should this just be Amplicon? The AmpliconStep description might be too technical, what do you think?

qp_klp/AmpliconStep.py
```diff
@@ -0,0 +1,102 @@
...
+                             f"pipeline of type '{pipeline.pipeline_type}'")
+
+    def convert_bcl_to_fastq(self):
+        config = self.pipeline.configuration['bcl2fastq']
```
antgonza, line 20: Is this hardcoded? Just asking but if is maybe adding a note about it would be good.

qp_klp/AmpliconStep.py
```diff
@@ -0,0 +1,102 @@
...
+        # note that pipeline.sample_sheet in this case
```
antgonza, line 21: This note makes sense but it would be better if it was clear that this is by design and historically (EMP primers)

qp_klp/AmpliconStep.py
```diff
@@ -0,0 +1,102 @@
...
+        # is the dummy sample-sheet created by Pipeline.
+        super()._convert_bcl_to_fastq(config, self.pipeline.sample_sheet)
+
+    def quality_control(self):
```
antgonza, line 25: Something that concerns me about this is that this looks like it's completely different from other pipelines, is that right? I thought the idea was to merge it. Note that I haven't read any code below.

qp_klp/AmpliconStep.py
```diff
@@ -0,0 +1,102 @@
...
+            job_output = [x for x in job_output if x.endswith('fastq.gz')]
+            job_output = [x for x in job_output if
+                          not basename(x).startswith('Undetermined')]
+
+            for raw_fastq_file in job_output:
+                new_path = join(faked_output_folder, basename(raw_fastq_file))
+                copyfile(raw_fastq_file, new_path)
+
+    def generate_reports(self, config, input_file_path):
+        config = self.pipeline.configuration['fastqc']
```
antgonza, line 81: Should this be part of self/super() and not have this here?

qp_klp/AmpliconStep.py
```diff
@@ -0,0 +1,102 @@
...
+        super()._generate_reports(config, self.pipeline.mapping_file)
+
+        return None  # amplicon doesn't need project names
+
+    def generate_prep_file(self):
```
antgonza, line 86: Same as comment above, this feels like the other pipeline is going to have the same methods

qp_klp/AmpliconStep.py
```diff
@@ -0,0 +1,102 @@
...
+        seqpro_path = config['seqpro_path'].replace('seqpro', 'seqpro_mf')
+
+        job = super()._generate_prep_file(config,
+                                          self.pipeline.mapping_file,
+                                          seqpro_path,
+                                          self.project_names)
+
+        self.prep_file_paths = job.prep_file_paths
+
+    def generate_commands(self):
```
antgonza, line 98: Same here, this should be part of the main class, no?

qp_klp/MetagenomicStep.py
```diff
@@ -0,0 +1,159 @@
...
+                             f"pipeline of type '{pipeline.pipeline_type}'")
+
+        # Note: FailedSamplesRecord is not used when processing amplicon as the
+        # samples are processed as a single fastq file and hence that info
+        # is not available.
+        self.fsr = FailedSamplesRecord(self.pipeline.output_path,
+                                       pipeline.sample_sheet.samples)
+        self.project_names = None
+
+    def convert_bcl_to_fastq(self):
```
antgonza, line 28: Yup, this looks like duplicated code between the two classes.

antgonza: Going to skip to the next section.

qp_klp/klp.py
```diff
@@ -40,66 +45,180 @@ def sequence_processing_pipeline(qclient, job_id, parameters, out_dir):
...
+
+        step.convert_bcl_to_fastq()
+
+        step.quality_control()
+
+        step.generate_reports()
+
+        step.generate_prep_file()
+
+        from_qiita = {}
```
antgonza, line 184: Looking at this and the code in Step is clear that this is not a step but a controller, similar to what you used to have of Pipeline; is that right? If yes, why the rename?

qp_klp/tests/test_amplicon_step.py
```diff
@@ -0,0 +1,65 @@
...
+from os.path import join, abspath
+from functools import partial
+from os import makedirs
+import json
+
+
+class AmpliconStepTests(TestCase):
+    def setUp(self):
+        package_root = abspath('./qp_klp')
+        self.path = partial(join, package_root, 'tests', 'data')
```
antgonza, line 20: why is this part of the object if it's only use within the setUp? Is the plan to reuse it somewhere else?

qp_klp/tests/test_amplicon_step.py
```diff
@@ -0,0 +1,65 @@
...
+
+    def test_creation(self):
+        # Test base-class creation method, even though base-class will never
+        # be instantiated by itself in normal usage.
+
+        # dummy used for constructor testing.
+        sn_tid_map_by_project = {}
+
+        # create metagenomic pipeline for failure tests.
+        metagenomic_pipeline = Pipeline(self.good_config_file,
```
antgonza, line 39: metagenomic_pipeline? ?

qp_klp/tests/test_amplicon_step.py
```diff
@@ -0,0 +1,65 @@
...
+                                        'metagenomic',
+                                        None)
+
+        with self.assertRaisesRegex(ValueError, "A pipeline object is needed"
+                                                " to initialize Step"):
+            AmpliconStep(None, self.qiita_id, sn_tid_map_by_project, None)
+
+        with self.assertRaisesRegex(ValueError, "A Qiita job-id is needed to "
+                                                "initialize Step"):
+            AmpliconStep(metagenomic_pipeline, None,
```
antgonza, line 54: `AmpliconStep(metagenomic_pipeline`? It can't be both, right?

antgonza (changes requested):
Thank you @charles-cowart, I have some concerns about the possibility of duplication of code; please let me know what you think.

.gitignore
charles-cowart: For this refactor I added the configuration.json file that is also a part of the mg-scripts repo into this repo. This necessitated removing this line from .gitignore so it could be added. This information is needed for running tests. In the past we were able to avoid adding configuration.json by hardcoding the data-structure in the tests. I wanted to avoid this in the new tests.

qp_klp/AmpliconStep.py
charles-cowart: Your call. It's more wordy but I think it was convention for so long to name subclasses like this that I got used to it. Amplicon would be nice and shorter too; not sure if it will get confusing with other usages of the word 'amplicon' in the code. What do you propose?

qp_klp/AmpliconStep.py
charles-cowart: No less than previously in qp-klp but yeah, the knowledge to look for 'bcl2fastq' for Amplicon and 'bcl-convert' for Metagenomics is hardcoded. I can think about how to do that in a more thoughtful way, or listen to suggestions. For now I'll add a comment.

qp_klp/tests/test_amplicon_step.py
charles-cowart: I am indeed cheating a little. I'm testing that missing parameters raise the right Errors and I know that due to the ordering of checks, it doesn't matter whether I use a metagenomic pipeline or an amplicon pipeline. I'll use an amplicon pipeline to avoid confusion.

charles-cowart:
@antgonza I've pushed a commit of quick updates based on your feedback, thanks! I haven't responded to all of your concerns yet, but I will shortly.

qp_klp/AmpliconStep.py
charles-cowart: The Jobs classes encapsulate all of the functionality we saw in the original shell scripts along with things we expected we'd want to have at that time. Since the actual creation and use of the Jobs classes was in qp-klp, it became the natural place to do work needed for each step and stick in additional features. A good example of this is generating failed sample records. We added it to the metagenomic pipeline but we removed it for the amplicon pipeline since we don't demultiplex the sample-names until the data is passed to Qiita.

Hence, each version of the pipeline has something they need to do at each stage; many of the stages share many of the same steps, but some need to do things before or after. In a few cases such as the QC stage of the amplicon pipeline, we need to skip adapter trimming and human filtering entirely and instead move and rename raw fastq files so they're in the form and location expected by the FastQC step.

AmpliconStep, MetagenomicStep, and future Step classes each encapsulate all the work that needs to be done at each stage. The elements common to each version are moved into the base class and called by the child classes instead of having the same code duplicated in each version.

In some cases where there's no difference between the two pipelines, they'll both just call the base class version of the step. In many cases, there'll be work that needs to be done or after the base call like generating failed sample reports in the metagenomics pipeline or setting up the ConvertJob object to use bcl2fastq instead of bcl-convert and creating a dummy sample-sheet for amplicon.

That's perhaps why the files look similar, even though there isn't much duplication of actual lines of code. The reason why the base class versions of the functions begin with '_' e.g. _quality_control() is because often the child versions have a different set of parameters and Python requires overriding methods to have the same parameter list. I could get around that with kwarg but I personally like the clarity of position parameters.

.gitignore
antgonza: Does this mean that now we have 2 config files one provided by mg-scripts and another by this repo?

qp_klp/AmpliconStep.py
antgonza: Just to be clear, I'm talking about the ValueError message, my concern is that a user will not know the difference between AmpliconStep/Amplicon and might confuse them, what do you think?

qp_klp/AmpliconStep.py
antgonza: Just to be clear, the point of this PR/work is to make this repo more python like (including nicer tests and remove duplication of code) vs. keeping the structure of the original shell script. 

Anyway, I understand your line of thought here: an object for each "Step" and have shared code in the base clase; which I guess is fine. Thank you for the explanation.

qp_klp/tests/test_amplicon_step.py
antgonza: cheating a little or a lot, sound wrong, right? Thank you for making the change.

antgonza (commented):
Thank you for the comments before. Now, a general question, if you are "extracting" the general functions from Amplicon/Metagenomic Step and adding them to the base class Step, does it really make sense to have a klp_util.py file? I mean basically you have to access method X in Amplicon/Metagenomic Step, that will call X method in Step and then do X in klp_util.py; no? Would it be better to remove one step and just go either Amplicon/Metagenomic Step -> klp_util.py or Amplicon/Metagenomic Step -> Step and remove klp_util.py?

.gitignore
charles-cowart: Yes. The alternative is to keep the data-structure as a const in the test code.

qp_klp/AmpliconStep.py
charles-cowart: I see where you're coming from but it could be argued that Error messages are as much for the developer as the user. The stack-traces we show in Qiita to the user have little value to the user, but convey that an error occurred and provide valuable insight to the developer.

qp_klp/AmpliconStep.py
charles-cowart: Moved into the base _generate_reports() class. Thanks!

qp_klp/AmpliconStep.py
charles-cowart: Not in this case, as we want to change the binary name for Amplicon runs from 'seqpro' to 'seqpro_mf'.

qp_klp/AmpliconStep.py
charles-cowart: Not sure if you mean the method above generate_commands() or generate_commands() itself, but in both cases each version needs different things done after the base call. In the case of generate_commands(), 'ConvertJob/Logs' needs to be included in the reports-ConvertJob.tgz tarball, whereas that directory is not created by bcl2fastq and therefore the Amplicon version.

qp_klp/MetagenomicStep.py
charles-cowart: In the case of convert_bcl_to_fastq(), one uses configuration info for bcl2fastq while the other uses bcl-convert. Also the metagenomic version generates a failed sample report afterward. Not sure where you're seeing the duplication.

qp_klp/tests/test_amplicon_step.py
charles-cowart: Good catch! No reason, just an artifact of having been cut and pasted from older code.

qp_klp/klp.py
charles-cowart: Not like Pipeline. 

> Thank you for the comments before. Now, a general question, if you are "extracting" the general functions from Amplicon/Metagenomic Step and adding them to the base class Step, does it really make sense to have a klp_util.py file? I mean basically you have to access method X in Amplicon/Metagenomic Step, that will call X method in Step and then do X in klp_util.py; no? Would it be better to remove one step and just go either Amplicon/Metagenomic Step -> klp_util.py or Amplicon/Metagenomic Step -> Step and remove klp_util.py?

We're definitely on the same page. The idea was to incorporate klp_util.py into the base Step class. I left them as is for the time being, just because so many other things had changed already.

charles-cowart:
> Thank you for the comments before. Now, a general question, if you are "extracting" the general functions from Amplicon/Metagenomic Step and adding them to the base class Step, does it really make sense to have a klp_util.py file? I mean basically you have to access method X in Amplicon/Metagenomic Step, that will call X method in Step and then do X in klp_util.py; no? Would it be better to remove one step and just go either Amplicon/Metagenomic Step -> klp_util.py or Amplicon/Metagenomic Step -> Step and remove klp_util.py?

We're definitely on the same page. The idea was to incorporate klp_util.py into the base Step class. I left them as is for the time being, just because so many other things had changed already.

antgonza:
> We're definitely on the same page. The idea was to incorporate klp_util.py into the base Step class. I left them as is for the time being, just because so many other things had changed already.

I think this will be the only blocker before merging this initial PR and this would be the base for future ones, could you do this now before waiting for a better time?

qp_klp/Amplicon.py
```diff
@@ -0,0 +1,105 @@
...
+                new_path = join(faked_output_folder, basename(fastq_file))
+                copyfile(fastq_file, new_path)
+
+            # FastQC expects the ConvertJob output to also be organized by
+            # project. Since this would entail running the same ConvertJob
+            # multiple times on the same input with just a name-change in
+            # the dummy sample-sheet, we instead perform ConvertJob once
+            # and copy the output from ConvertJob's output folder into
+            # ConvertJob's output folder/project1, project2 ... projectN.
+            faked_output_folder = join(raw_fastq_files_path, project_name)
```
antgonza, line 72: Not a fan of this name, why fake?

qp_klp/Amplicon.py
```diff
@@ -0,0 +1,105 @@
...
+            makedirs(faked_output_folder)
+            job_output = [join(raw_fastq_files_path, x) for x in
+                          listdir(raw_fastq_files_path)]
+            job_output = [x for x in job_output if isfile(x)]
+            job_output = [x for x in job_output if x.endswith('fastq.gz')]
+            job_output = [x for x in job_output if
```
antgonza, line 78: Why not join all these if statements and only loop the list once?

qp_klp/Step.py
```diff
@@ -0,0 +1,432 @@
...
+        # to reorganize the running-results and write them out to disk.
+
+        self.output_path = join(output_dir, 'failed_samples.html')
+
+        # create an initial dictionary with sample-ids as keys and their
+        # associated project-name and status as values. Afterwards, we'll
+        # filter out the sample-ids w/no status (meaning they were
+        # successfully processed) before writing the failed entries out to
+        # file.
+        self.failed = {x.Sample_ID: [x.Sample_Project, None] for x in samples}
```
antgonza, line 31: is this really failed? At this stage it looks like all samples; it might be confusing for other devs.

antgonza: One option is to have: self.samples (all samples) and then another one with self.samples_used and self.failed can be the difference between both.

qp_klp/Metagenomic.py
```diff
@@ -0,0 +1,161 @@
...
+
+class Metagenomic(Step):
+    def __init__(self, pipeline, master_qiita_job_id, sn_tid_map_by_project,
+                 status_update_callback=None):
+        super().__init__(pipeline,
+                         master_qiita_job_id,
+                         sn_tid_map_by_project,
+                         status_update_callback)
+
+        if pipeline.pipeline_type != 'metagenomic':
```
antgonza, line 16: Should this be metatranscriptomic too? Note that this might be concerning as there are several places where this check is done, perhaps having a global set or something might be better; it could be part of the object too.

qp_klp/Metagenomic.py
```diff
@@ -0,0 +1,161 @@
...
+                # move uncompressed sifs to upload_dir.
+                tmp = f'cd {out_dir}; mv {blanks_file} {upload_dir}'
+                self.cmds.append(tmp)
+
+            # record that something is being moved into a Qiita Study.
+            # this will allow us to notify the user which Studies to
+            # review upon completion.
+            touched_studies.append((qiita_id, project))
+
+            if self.pipeline.pipeline_type == 'metagenomic':
```
antgonza, line 106: This is another case about Metatranscriptomic

qp_klp/klp.py
```diff
@@ -41,65 +73,178 @@ def sequence_processing_pipeline(qclient, job_id, parameters, out_dir):
...
+        if errors:
+            raise PipelineError('\n'.join(errors))
+
+        # find the uploads directory all trimmed files will need to be
+        # moved to and generate a map.
+        special_map = Step.generate_special_map(
+            qclient.get("/qiita_db/artifacts/types/"),
+            pipeline.get_project_info())
+
+        if pipeline.pipeline_type == 'metagenomic':
```
antgonza, line 185: metatransciptomic??

antgonza (commented):
Some questions.

qp_klp/Amplicon.py
charles-cowart: It's faked in the sense that Amplicon doesn't invoke QCJob but the subsequent step expects a QCJob style layout. Rather than alter QCJob we simply copy the raw output from ConvertJob into something that looks like QCJob output.

charles-cowart: I'll come up with an alternative name.

antgonza: I think output_folder is enough, no?

charles-cowart: same. just changed it to that, actually.

qp_klp/Step.py
charles-cowart: I changed the name from self.failed to self.sample_state - you're right, we're initializing the member with the entire list of samples but we don't know that any of them are failed yet.

qp_klp/tests/test_amplicon_step.py
```diff
@@ -0,0 +1,64 @@
...
+        super().setUp()
+        self.good_mapping_file = join(abspath('./qp_klp'), 'tests', 'data',
+                                      'good-mapping-file.txt')
+
+    def test_creation(self):
+        # Test base-class creation method, even though base-class will never
+        # be instantiated by itself in normal usage.
+
+        # dummy used for constructor testing.
+        sn_tid_map_by_project = {}
```
antgonza, line 25: Do you really need a variable for this? Can the `{}` be passed in the function?

antgonza (commented):
A few more comments.

antgonza (commented):
A few more comments.

qp_klp/Amplicon.py
```diff
@@ -0,0 +1,103 @@
...
+            # multiple times on the same input with just a name-change in
+            # the dummy sample-sheet, we instead perform ConvertJob once
+            # and copy the output from ConvertJob's output folder into
+            # ConvertJob's output folder/project1, project2 ... projectN.
+            output_folder = join(raw_fastq_files_path, project_name)
+            makedirs(output_folder)
+            job_output = [join(raw_fastq_files_path, x) for x in
+                          listdir(raw_fastq_files_path)]
+            job_output = [x for x in job_output if isfile(x) and x.endswith(
+                'fastq.gz') and not basename(x).startswith('Undetermined')]
```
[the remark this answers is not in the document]
charles-cowart: I think it might be better to keep the separate, because you want x in isfile(x) in the second one to be the full path created by join(raw_fastq_files_path, x) in the first one and if we merged them that would mean we'd have something like:

job_output = [join(raw_fastq_files_path, x) for x in listdir(raw_fastq_files_path) if isfile(join(raw_fastq_files_path, x)) and x.endswith('fastq.gz') and not basename(x).startswith('Undetermined')]]

The join() has to be present twice.

qp_klp/Amplicon.py
```diff
@@ -0,0 +1,103 @@
...
+                copyfile(raw_fastq_file, new_path)
+
+    def generate_reports(self, input_file_path):
+        super()._generate_reports(self.pipeline.mapping_file)
+        return None  # amplicon doesn't need project names
+
+    def generate_prep_file(self):
+        config = self.pipeline.configuration['seqpro']
+
+        seqpro_path = config['seqpro_path'].replace('seqpro', 'seqpro_mf')
```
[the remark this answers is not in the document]
charles-cowart: It's an easy enough change on this side of things, and a little bit of refactoring for metapool package. I can make the change by the end of the week in between other things.

qp_klp/Metagenomic.py
```diff
@@ -0,0 +1,161 @@
...
+    def convert_bcl_to_fastq(self):
+        # The 'bcl-convert' key is a convention hard-coded into mg-scripts and
+        # qp-klp projects. Currently meta*omic jobs use bcl-convert for its
+        # improved performance over bcl2fastq. The name and path of the
+        # executable, the resource requirements to instantiate a SLURM job
+        # with, etc. are stored in configuration['bcl-convert''].
+        config = self.pipeline.configuration['bcl-convert']
+        job = super()._convert_bcl_to_fastq(config,
+                                            self.pipeline.sample_sheet.path)
+        self.fsr.write(job.audit(self.pipeline.get_sample_ids()), 'ConvertJob')
```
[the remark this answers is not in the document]
charles-cowart: Sure, no problem!

PR #61, opened 2023-05-24
Testing updates

charles-cowart:
This PR should require https://github.com/biocore/mg-scripts/pull/98 to be approved and merged beforehand to pass CI.

### DIFF
--- a/qp_klp/Amplicon.py
+++ b/qp_klp/Amplicon.py
@@ -41,8 +41,11 @@ class Amplicon(Step):
             output_folder = join(self.pipeline.output_path,
                                  'QCJob',
                                  project_name,
-                                 # for legacy purposes, do not replace this
-                                 # with AMPLICON_TYPE
+                                 # for legacy purposes, output folders are
+                                 # either 'trimmed_sequences', 'amplicon', or
+                                 # 'filtered_sequences'. Hence, this folder
+                                 # is not defined using AMPLICON_TYPE as that
+                                 # value may or may not equal the needed value.
                                  'amplicon')
 
             makedirs(output_folder)
@@ -87,30 +90,46 @@ class Amplicon(Step):
         super()._generate_reports()
         return None  # amplicon doesn't need project names
 
+    def _get_data_type(self, prep_file_path):
+        metadata = Step.parse_prep_file(prep_file_path)
+        if 'target_gene' in metadata.columns:
+            tg = set(metadata['sample target_gene'])
+            if len(tg) != 1:
+                raise ValueError("More than one value for target_gene")
+            else:
+                tg = tg.pop()
+
+            data_type = None
+            for key in Step.AMPLICON_SUB_TYPES:
+                # the key sub_types are often substrings of tg,
+                # rather than equal to tg.
+                if key in tg:
+                    data_type = key
+
+            if data_type is None:
+                raise ValueError(f"Valid data-types "
+                                 f"{Step.AMPLICON_SUB_TYPES} could not"
+                                 " be found within target_gene value "
+                                 f"'{tg}'")
+
+            return data_type
+        else:
+            raise ValueError("'target_gene' column not present in "
+                             "generated prep-files")
+
     def generate_touched_studies(self, qclient):
+        results = {}
         for study_id in self.prep_file_paths:
             for prep_file_path in self.prep_file_paths[study_id]:
-                metadata = Step.parse_prep_file(prep_file_path)
-                if 'target_gene' in metadata[list(metadata.keys())[0]]:
-                    tg = metadata[list(metadata.keys())[0]]['target_gene']
-                    data_type = None
-                    for key in Step.AMPLICON_SUB_TYPES:
-                        if key in tg:
-                            data_type = key
+                results[prep_file_path] = self._get_data_type(prep_file_path)
 
-                    if data_type is None:
-                        raise ValueError("A valid data-type could not be "
-                                         "derived from target_gene column")
-
-            super()._generate_touched_studies(qclient, data_type)
+        super()._generate_touched_studies(qclient, results)
 
     def generate_prep_file(self):
         config = self.pipeline.configuration['seqpro']
-
         seqpro_path = config['seqpro_path'].replace('seqpro', 'seqpro_mf')
-
-        projects = self.pipeline.get_project_info()
-        project_names = [x['project_name'] for x in projects]
+        project_names = [x['project_name'] for x in
+                         self.pipeline.get_project_info()]
 
         job = super()._generate_prep_file(config,
                                           self.pipeline.mapping_file_path,
--- a/qp_klp/Amplicon.py
+++ b/qp_klp/Amplicon.py
@@ -92,25 +92,17 @@ class Amplicon(Step):
     def _get_data_type(self, prep_file_path):
         metadata = Step.parse_prep_file(prep_file_path)
         if 'target_gene' in metadata.columns:
-            tg = set(metadata['sample target_gene'])
+            # remove duplicate values, then convert back to list for
+            # accession.
+            tg = list(set(metadata['sample target_gene']))
             if len(tg) != 1:
                 raise ValueError("More than one value for target_gene")
 
-            data_type = None
-            for key in Step.AMPLICON_SUB_TYPES:
-                # the key sub_types are often substrings of tg,
-                # rather than equal to tg.
-                if key in tg:
-                    data_type = key
-                    break
+            if tg[0] in Step.AMPLICON_SUB_TYPES:
+                return tg[0]
 
-            if data_type is None:
-                raise ValueError(f"Valid data-types "
-                                 f"{Step.AMPLICON_SUB_TYPES} could not"
-                                 " be found within target_gene value "
-                                 f"'{tg.pop()}'")
-
-            return data_type
+            raise ValueError(f"'{tg[0]}' is not a valid type - valid data-"
+                             f"types are {Step.AMPLICON_SUB_TYPES}")
         else:
             raise ValueError("'target_gene' column not present in "
                              "generated prep-files")
--- a/qp_klp/Amplicon.py
+++ b/qp_klp/Amplicon.py
@@ -95,8 +95,6 @@ class Amplicon(Step):
             tg = set(metadata['sample target_gene'])
             if len(tg) != 1:
                 raise ValueError("More than one value for target_gene")
-            else:
-                tg = tg.pop()
 
             data_type = None
             for key in Step.AMPLICON_SUB_TYPES:
@@ -109,7 +107,7 @@ class Amplicon(Step):
                 raise ValueError(f"Valid data-types "
                                  f"{Step.AMPLICON_SUB_TYPES} could not"
                                  " be found within target_gene value "
-                                 f"'{tg}'")
+                                 f"'{tg.pop()}'")
 
             return data_type
         else:
@@ -118,9 +116,9 @@ class Amplicon(Step):
 
     def generate_touched_studies(self, qclient):
         results = {}
-        for study_id in self.prep_file_paths:
-            for prep_file_path in self.prep_file_paths[study_id]:
-                results[prep_file_path] = self._get_data_type(prep_file_path)
+        for study_id, pf_paths in self.prep_file_paths.items():
+            for pf_path in pf_paths:
+                results[pf_path] = self._get_data_type(pf_path)
 
         super()._generate_touched_studies(qclient, results)
 
--- a/qp_klp/Amplicon.py
+++ b/qp_klp/Amplicon.py
@@ -41,7 +41,9 @@ class Amplicon(Step):
             output_folder = join(self.pipeline.output_path,
                                  'QCJob',
                                  project_name,
-                                 Step.AMPLICON_TYPE)
+                                 # for legacy purposes, do not replace this
+                                 # with AMPLICON_TYPE
+                                 'amplicon')
 
             makedirs(output_folder)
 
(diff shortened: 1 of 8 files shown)


qp_klp/Amplicon.py
```diff
@@ -41,7 +41,9 @@ def quality_control(self):
             output_folder = join(self.pipeline.output_path,
                                  'QCJob',
                                  project_name,
-                                 Step.AMPLICON_TYPE)
+                                 # for legacy purposes, do not replace this
```
antgonza, line 44: Can you add more info about this? With the limited text is not clear why this is important.

qp_klp/Amplicon.py
```diff
@@ -81,23 +83,43 @@ def quality_control(self):
...
+                metadata = Step.parse_prep_file(prep_file_path)
+                if 'target_gene' in metadata[list(metadata.keys())[0]]:
+                    tg = metadata[list(metadata.keys())[0]]['target_gene']
+                    data_type = None
+                    for key in Step.AMPLICON_SUB_TYPES:
+                        if key in tg:
+                            data_type = key
+
+                    if data_type is None:
+                        raise ValueError("A valid data-type could not be "
```
antgonza, line 102: Can you add the wrong value and the valid values so users can fix?

qp_klp/Amplicon.py
```diff
@@ -81,23 +83,43 @@ def quality_control(self):
...
+            for prep_file_path in self.prep_file_paths[study_id]:
+                metadata = Step.parse_prep_file(prep_file_path)
+                if 'target_gene' in metadata[list(metadata.keys())[0]]:
+                    tg = metadata[list(metadata.keys())[0]]['target_gene']
+                    data_type = None
+                    for key in Step.AMPLICON_SUB_TYPES:
+                        if key in tg:
+                            data_type = key
+
+                    if data_type is None:
```
antgonza, line 101: This code seems to assume that `['target_gene']` can have multiple values in a single prep, right? However, it should only be one; if there are multiple that should be an error.

qp_klp/Amplicon.py
```diff
@@ -81,23 +83,43 @@ def quality_control(self):
...
+
+            super()._generate_touched_studies(qclient, data_type)
+
     def generate_prep_file(self):
         config = self.pipeline.configuration['seqpro']
 
         seqpro_path = config['seqpro_path'].replace('seqpro', 'seqpro_mf')
 
+        projects = self.pipeline.get_project_info()
+        project_names = [x['project_name'] for x in projects]
```
antgonza, line 113: Can you merge these 2 lines in a single loop?
`project_names = [x['project_name'] for x in project_names = [x['project_name'] for x in projects]]`

qp_klp/Step.py
```diff
@@ -369,32 +379,149 @@ def generate_sifs(self, from_qiita):
...
-
-    def update_blanks_in_qiita(self, sifs, qclient):
-        for sif_path in sifs:
+    def get_tube_ids_from_qiita(self, qclient):
+        # Update get_project_info() so that it can return a list of
+        # samples in projects['samples']. Include blanks in projects['blanks']
+        projects = self.pipeline.get_project_info(short_names=True)
+
+        # just in case there are duplicate qiita_ids
+        qiita_ids = list(set([x['qiita_id'] for x in projects]))
```
antgonza, line 388: These 2 lines can also be merged.

qp_klp/Step.py
```diff
@@ -369,32 +379,149 @@ def generate_sifs(self, from_qiita):
...
+        # called but no tube-ids were found for any project.
+        self.tube_id_map = results
+        self.samples_in_qiita = results2
+
+        # not needed, but useful in displaying debugging info
+        return results
+
+    def donald_duck(self):
+        projects = self.pipeline.get_project_info(short_names=True)
+        self.foo("PROJECTS: %s" % projects)
```
antgonza, line 422: ????

qp_klp/klp.py
```diff
@@ -155,63 +155,44 @@ def sequence_processing_pipeline(qclient, job_id, parameters, out_dir):
...
-            samples = set(pipeline.get_sample_ids())
+        status_line.update_current_message()
 
-            # strip any leading zeroes from the sample-ids. Note that
-            # if a sample-id has more than one leading zero, all of
-            # them will be removed.
-            # my_samples = {x.lstrip('0') for x in samples}
+        # corroborate sample-ids/tube-ids in sample-sheet or mapping file
+        # against what's in Qiita.
+        blah = step.donald_duck()
```
antgonza, line 170: ???

antgonza (changes requested):
A few requests and questions.

qp_klp/Amplicon.py
charles-cowart: We do assume that there will only be one type, but what I've noticed is that the value is often not '16S', '18S', or 'ITS' but a value where one of those will be a substring within. Hence the need to loop through the enumerated types and check.

qp_klp/Step.py
charles-cowart: Thanks for the catch! Yeah I probably should change that working name...

qp_klp/Amplicon.py
```diff
@@ -81,23 +86,59 @@ def quality_control(self):
...
+        super()._generate_reports()
         return None  # amplicon doesn't need project names
 
+    def _get_data_type(self, prep_file_path):
+        metadata = Step.parse_prep_file(prep_file_path)
+        if 'target_gene' in metadata.columns:
+            tg = set(metadata['sample target_gene'])
+            if len(tg) != 1:
+                raise ValueError("More than one value for target_gene")
+            else:
```
antgonza, line 99: Do you really need the else here?

qp_klp/Amplicon.py
```diff
@@ -81,23 +86,59 @@ def quality_control(self):
...
+
+            return data_type
+        else:
+            raise ValueError("'target_gene' column not present in "
+                             "generated prep-files")
+
+    def generate_touched_studies(self, qclient):
+        results = {}
+        for study_id in self.prep_file_paths:
+            for prep_file_path in self.prep_file_paths[study_id]:
```
antgonza, line 123: This is a dictionary? It looks like the iter is doing that, so why not use items()?

qp_klp/Metagenomic.py
```diff
@@ -59,8 +61,18 @@ def generate_prep_file(self):
 
         self.prep_file_paths = job.prep_file_paths
 
-    def generate_commands(self, special_map, server_url,
-                          touched_studies_prep_info):
+    def generate_touched_studies(self, qclient):
+        results = {}
+        for study_id in self.prep_file_paths:
+            for prep_file_path in self.prep_file_paths[study_id]:
```
antgonza, line 67: Similar dict/items?

qp_klp/Step.py
```diff
@@ -369,32 +379,141 @@ def generate_sifs(self, from_qiita):
...
-                                    df.at[i, "sample_name"] = tube_id
-                        df.to_csv(prep_file, index=False, sep="\t")
-
-    def update_blanks_in_qiita(self, sifs, qclient):
-        for sif_path in sifs:
+    def get_tube_ids_from_qiita(self, qclient):
+        # Update get_project_info() so that it can return a list of
+        # samples in projects['samples']. Include blanks in projects['blanks']
+        # just in case there are duplicate qiita_ids
+        qiita_ids = list(set([x['qiita_id'] for x in
```
antgonza, line 386: Why not create the set from the beginning? 
qiita_ids = list({x['qiita_id'] for x in ... })

antgonza (commented):
A few more comments.

qp_klp/Amplicon.py
charles-cowart: IMHO, sanity checks can be important. In this case there are plenty of legacy prep-info-files out there w/out target_gene as a column. Granted, they might not pass through this code but I think it's best to raise the Error if a value can't be found, rather than continue and output an unpredictable file.

qp_klp/Amplicon.py
charles-cowart: As far as I understand it, I could call self.prep_file_paths.items() and get back a (k,v) pair, but the v in this case will still be a list of prep-file-paths that I need to loop through. I'm not sure what the gain is.

qp_klp/Step.py
charles-cowart: It was an insurance measure for uniqueness during development, in case multiple projects with the same qiita-id were declared in a sample-sheet or pre-prep file. I removed the list() and set() casts entirely.

qp_klp/Metagenomic.py
charles-cowart: See previous comment.

qp_klp/Amplicon.py
antgonza: Right, note that my question wasn't the if but the else ... I do not think is necessary.

qp_klp/Amplicon.py
antgonza: That you don't need to index the dictionary twice! Once to get the keys and another to retrieve the value of the key ...

qp_klp/Metagenomic.py
antgonza: See previous comment.

qp_klp/Amplicon.py
```diff
@@ -81,23 +85,57 @@ def quality_control(self):
...
+            tg = set(metadata['sample target_gene'])
+            if len(tg) != 1:
+                raise ValueError("More than one value for target_gene")
+
+            data_type = None
+            for key in Step.AMPLICON_SUB_TYPES:
+                # the key sub_types are often substrings of tg,
+                # rather than equal to tg.
+                if key in tg:
+                    data_type = key
```
antgonza, line 104: should this have a break so once found it doesn't continue?

qp_klp/Step.py
```diff
@@ -143,22 +149,27 @@ def parse_prep_file(cls, prep_file_path):
...
             upload_path = join(results['uploads'], project['qiita_id'])
             makedirs(upload_path, exist_ok=True)
             special_map.append((project['project_name'], upload_path,
                                 project['qiita_id']))
 
-        return special_map
+        self.special_map = special_map
+
+        # for ease of debugging
+        return self.special_map
```
antgonza, line 172: if this is adding `self.special_map` do we really need the return? Also, should this be an internal method (_ generate_special_map vs. generate_special_map)?

qp_klp/Step.py
```diff
@@ -369,32 +386,140 @@ def generate_sifs(self, from_qiita):
...
+                # and if prep-files have empty sample-names we'll know.
+                tids_by_qiita_id[str(qiita_id)] = tids
+
+        # use empty dict {} as an indication that get_tube_ids_from_qiita was
+        # called but no tube-ids were found for any project.
+        self.tube_id_map = tids_by_qiita_id
+        self.samples_in_qiita = sample_names_by_qiita_id
+
+        # not needed, but useful in displaying debugging info
+        return tids_by_qiita_id
```
antgonza, line 424: If not needed, please rm (similar to my previous comment) ... you can always use self.tube_id_map for debugging, right?

qp_klp/Step.py
```diff
@@ -369,32 +386,140 @@ def generate_sifs(self, from_qiita):
...
+                continue
+
+            # remove leading zeroes if they exist to match Qiita results.
+            sample_name = sample_name.lstrip('0')
+            if sample_name in tube_id_map[qiita_id]:
+                df.at[i, "sample_name"] = tube_id_map[qiita_id][sample_name]
+
+        df.to_csv(prep_file_path, index=False, sep="\t")
+
+    def _overwrite_prep_files(self, prep_file_paths):
```
antgonza, line 488: The classmethod _overwrite_prep_file and this one are really similar, maybe select something less similar to avoid confusion? Also, it looks like the only place where _overwrite_prep_file is called is within _overwrite_prep_files; do we really need that split?

qp_klp/tests/test_step.py
```diff
@@ -108,10 +244,10 @@ def setUp(self):
 
     def _is_writable(self, a_path):
```
antgonza, line 245: Why not use [os.access](https://www.tutorialspoint.com/python/os_access.htm)?

antgonza (commented):
More comments; getting there.

qp_klp/Step.py
charles-cowart: I like it because it makes the code in test more obvious:
_obs = step.generate_special_map(fake_client)_
but I can change it to:
_step.generate_special_map(fake_client)
obs = step.special_map_

qp_klp/Step.py
charles-cowart: I can change the names to be more unique but I'd like to keep it as two separate functions. I'll be writing an extra test for the static method.

qp_klp/Amplicon.py
charles-cowart: I think that's a good idea. However I don't think it's a bad idea to have this sanity check here for a little while as defensive programming.

wasade: I agree re defensive! If an exact match isnt found, could an exception be raised? I think that would be the most defensive posture

charles-cowart: From what I've seen, the values in target_gene aren't usually exact matches for {16S, 18S, ITS}, at least for 16S; that's why we search for them as substrings. We do raise an exception if we don't find any matches; do you think we should change to exact match? It's a new column (at least to me) so I'm not sure what to expect yet.

wasade: Yes, exact match is best. We know the specific target gene, and there any variation in name suggests something was manually filled in or modified upstream, and that leads to noise in Qiita

PR #68, opened 2023-06-16
Antonio and Charlie's updates from testing

charles-cowart:
Updates based on additional testing in qiita-rc.

### DIFF
--- a/qp_klp/Step.py
+++ b/qp_klp/Step.py
@@ -656,8 +656,8 @@ class Step:
                     # strip any leading zeroes from the sample-ids. Note that
                     # if a sample-id has more than one leading zero, all of
                     # them will be removed.
-                    foo = set([x.lstrip('0') for x in samples])
-                    not_in_qiita = foo - set(tids)
+                    not_in_qiita = set([x.lstrip('0') for x in samples]) - \
+                                   set(tids)
 
                 examples = tids[:5]
                 used_tids = True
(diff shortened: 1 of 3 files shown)


antgonza:
Does these changes need tests?

qp_klp/Step.py
```diff
@@ -634,18 +638,27 @@ def _compare_samples_against_qiita(self, qclient):
...
                 tids = [self.tube_id_map[qiita_id][sample] for sample in
                         self.tube_id_map[qiita_id]]
+
                 not_in_qiita = samples - set(tids)
+
+                if not_in_qiita:
+                    # strip any leading zeroes from the sample-ids. Note that
+                    # if a sample-id has more than one leading zero, all of
+                    # them will be removed.
+                    foo = set([x.lstrip('0') for x in samples])
```
antgonza, line 659: please use a better var name.

qp_klp/tests/test_step.py
```diff
@@ -146,6 +146,63 @@ def get(self, url):
         return None
 
 
+class AnotherFakeClient():
```
antgonza, line 149: Out of curiosity, how many fake clients? If many, can they be reused like:
```python
class FakeClientBase():
...
...

class FakeClient1(FakeClientBase)
...
...

class FakeClient2(FakeClientBase)
...
...
```

antgonza (commented):
Some questions.

qp_klp/tests/test_step.py
charles-cowart: There's just two. This new 'AnotherFakeClient()' just has a few specific samples to test how pre-check handles the situations we need it to. IMHO I don't think there's anything to be gained right now because most tests I think would use the first one, and they're really just wrappers to return dicts of metadata. If it became a reoccurring thing then perhaps, IMHO.

PR #70, opened 2023-08-09
metapool + wallclock time updates

charles-cowart:


charles-cowart:
@antgonza We will need these changes to qp-klp in addition to what was changed in mg-scripts.

antgonza:
I'll review this after #69 is merged.

charles-cowart:
@antgonza This should still be reviewed after #69 is merged.

antgonza:
@charles-cowart, conflicts after merging #69

charles-cowart:
@antgonza ready for review!

qp_klp/tests/data/good-sample-sheet.csv
antgonza, line 1: I'm not sure the changes in Well_description are correct, for example in the last sample `EP479894B04` to `desc_for_EP479894B04`; could you confirm this?

qp_klp/tests/data/good_sheet_w_replicates.csv
antgonza, line 1: It looks like this full file changed, is that correct? or another way to ask this, what changed in this file?

antgonza (commented):
A couple of questions.

qp_klp/tests/data/good-sample-sheet.csv
charles-cowart: I can revert it if that's your preference. I like to change legacy sample-name values in the well_description column when I encounter them as a form of defensive programming. We previously changed the code that looks for sample-names in the Well_description column; this change supports that.

qp_klp/tests/data/good_sheet_w_replicates.csv
charles-cowart: I'll confirm, one moment.

charles-cowart: The only meaningful change is the addition of the syndna_pool_number column in the [Data] section. However, since we've added a new column in the [Data] section, we had to add an additional ',' character to the end of each line in the other sections to preserve the sample-sheet specification. This will make every line appear as changed in any visual diff between the old and new.

Specifically, sample-sheets are in a tabular format with a uniform number of columns for each row, even though they can also be read like an INI file. The [Data] section usually determines the number of columns in a sample-sheet, since it has the largest number of required columns. This impacts the other sections by affecting the number of empty column values in their rows, captured by ',,...,'.

Since we added a column to the data section we had to add that column everywhere or else when the file is read into KLSampleSheet() and written back out, the output files won't be identical.

qp_klp/tests/data/good-sample-sheet.csv
antgonza: Asking the same question differently, is the prefix `desc_for_` the expected one? Asking because it looks weird to me, I guess because I never seen that prefix before.

charles-cowart: no it's one I made up for the file.

charles-cowart: I've gone ahead and removed the desc_for_ prefix for this file.

PR #72, opened 2023-08-17
Converted setup.py to pyproject.toml

charles-cowart:
plugin updated to use pyproject.toml rather than setup.py. This resolves the installation error occurring in CI and on fresh installations.

tests have not been updated yet to include changes from the latest versions of metapool and mg-scripts. 18 tests should still fail.

Scripts directory moved under qp_klp and files appended with .py suffix. 'cmd' scripts are generated for each command by pyproject.html and added to the vm's bin.

Other PRs in qp-knight-lab-processing should be reviewed and merged _after_ this PR is complete, approved, and merged.


charles-cowart:
@antgonza here is the first pass at migrating to pyproject.toml. No rush to review, just keeping you in the loop.

antgonza:
@charles-cowart note that tests are failing.

.github/workflows/qiita-plugin-ci.yml
```diff
@@ -154,5 +154,5 @@ jobs:
     - name: lint
       run: |
         pip install -q flake8
-        flake8 qp_klp setup.py scripts/*
+        flake8 qp_klp
```
antgonza, line 157: why `rm scripts/*`?

antgonza (changes requested):
Only one comment.

.github/workflows/qiita-plugin-ci.yml
charles-cowart: Pyproject.toml is now responsible for creating configure_klp and start_klp binaries. The syntax for creating them requires the contents of the scripts directory to now be part of the module (moved from being co-level with qp_klp directory to being a subdirectory of qp_klp). Since the scripts directory is now under qp_klp, it doesn't need to be tested separately.

antgonza: Got it, thank you for the explanation. Do you foresee any issues with `configure_klp` vs `configure_klp.py` ? 

BTW I'm still confused why this was necessary for this repo and if this is going to be needed for other repos ...

charles-cowart: There shouldn't be issues with configure_klp vs configure_klp.py, because the installer will generate a separate python script in 'bin' named 'configure_klp' that will call the same config() function in configure_klp.py. 

There isn't a way in pyproject.toml to install scripts the way setup.py does (scripts=['scripts/configure_klp', 'scripts/start_klp']). Instead, there's a section header [project.scripts] in pyproject.toml where you name entry points in your module and little wrapper scripts get created for them in bin.

charles-cowart: To be more explicit, we could take the entry point functions of configure_klp.py and start_klp.py move the them into the module, and remove these two files entirely and configure_klp and start_klp will still be created in bin.

PR #76, opened 2023-10-30
Adapting code & tests to use NuQCJob

charles-cowart:


qp_klp/Step.py
```diff
@@ -607,6 +613,58 @@ def load_preps_into_qiita(self, qclient):
 
         return df
 
+    def _load_preps_into_qiita2(self, qclient, prep_id, qiita_id, out_dir,
```
wasade, line 616: is `_load_preps_into_qiita()` still being used? If it is not, then suggest adopting that name. If it is, then the naming needs to be more appropriate

qp_klp/Step.py
```diff
@@ -607,6 +613,58 @@ def load_preps_into_qiita(self, qclient):
...
+                    f'{qiita_id}?prep_id={prep_id}')
+
+        if self.pipeline.pipeline_type == Step.AMPLICON_TYPE:
+            files = self._get_files_amplicon(out_dir, project)
+        else:
+            atype = 'per_sample_FASTQ'
+            files = self._get_files_meta(out_dir, project, prep_id)
+
+        for f_type in files:
+            if not files[f_type]:
```
wasade, line 629: it looks like this is a `dict`? if so then suggest:

```suggestion
        for f_type, size in files.items():
            if size == 0:
```

wasade: ...althoguh I don't understand the variable name `f_type`?

qp_klp/Step.py
```diff
@@ -607,6 +613,58 @@ def load_preps_into_qiita(self, qclient):
...
+                # if one or more of the expected list of reads is empty,
+                # raise an Error.
+                raise ValueError(f"'{f_type}' is empty")
+
+        # ideally we would use the email of the user that started the SPP
+        # run but at this point there is no easy way to retrieve it
+        pdata = {'user_email': 'qiita.help@gmail.com',
+                 'prep_id': prep_id,
+                 'artifact_type': atype,
```
wasade, line 638: It looks like this will produce a `NameError` if it's amplicon data

qp_klp/Step.py
```diff
@@ -607,6 +613,58 @@ def load_preps_into_qiita(self, qclient):
...
+                 'command_artifact_name': self.generated_artifact_name,
+                 'files': dumps(files)}
+
+        job_id = qclient.post('/qiita_db/artifact/', data=pdata)['job_id']
+
+        return {'Project': project, 'Qiita Study ID': qiita_id,
+                'Qiita Prep ID': prep_id, 'Qiita URL': surl,
+                'Prep URL': prep_url, 'Linking JobID': job_id}
+
+    def load_preps_into_qiita2(self, qclient):
```
wasade, line 648: ???

wasade (changes requested):
unit tests?

charles-cowart:
Still a work in progress.

charles-cowart:
Note this PR has been tested against https://github.com/biocore/mg-scripts/pull/110 and that PR should be merged before reviewing and merging this PR.

qp_klp/Step.py
charles-cowart: Thanks for pointing it out! This code is vestigial.

qp_klp/Step.py
charles-cowart: thanks! size == 0 is easier to read. I changed f_type to 'key' because that's really what they are - we're just checking the values associated w/each key to make sure they're not empty.

qp_klp/Step.py
charles-cowart: This became vestigial as well. I rewrote the code to handle both the plate_replicates case and without, across amplicon and meta*omics.

qp_klp/Step.py
charles-cowart: I reviewed the earlier versions of the code that handle loading for amplicon data and can't find a discrepancy in the parameters. Do you recall why you thought this? In either case I'll add a fresh amplicon test to the round of testing, once loading is re-enabled.

wasade: You previously used a pattern like:

```python
if foo == bar:
    atype = 'thing'
```

where the variable `atype` was only defined if a specific branch was taken, which would lead to a `NameError`

wasade: thank you for fixing it

charles-cowart: I think it was always like that but all good either way. Perhaps the rewrite to handle meta*omic and amplicon both w/replication and w/out took care of it. No worries. Ty for the review.

charles-cowart:
https://github.com/biocore/mg-scripts/pull/111 should be reviewed and merged first. GitHub reference should then be reverted back to normal. Then PR should be ready for review. The new test to validate NuQCJob script is an analog to the same test for legacy QCJob and should be sufficient.

charles-cowart:
Dependency https://github.com/biocore/mg-scripts/pull/112 should be reviewed and merged first before reviewing and merging this PR.

qp_klp/tests/test_step.py
```diff
@@ -271,6 +272,24 @@ class BaseStepTests(TestCase):
                 "job_pool_size": 30,
                 "job_max_array_length": 1000
             },
+            "nu-qc": {
+                "nodes": 1,
+                "nprocs": 16,
```
antgonza, line 277: Based on the other open PR, should this be connected/the-same as {{cores_per_task}}? If yes, could you make sure to use the same name, just for clarity?

antgonza (commented):
One question.

qp_klp/tests/test_step.py
charles-cowart: Good catch! I removed nprocs for NuQC because it's not being used and added cpus_per_task since it is being used. I added a comment briefly covering the difference between the two parameters after reading some literature.

antgonza (commented):
Looks good, only a couple of questions.

qp_klp/Step.py
```diff
@@ -280,36 +285,38 @@ def _convert_bcl_to_fastq(self, config, input_file_path):
 
         return convert_job
 
+    # CHARLIE
```
antgonza, line 288: rm?

charles-cowart: doh! Thanks for the catch!

PR #78, opened 2024-02-15
Updated to support changes in mg-scripts for multiple config.

charles-cowart:
qp-klp configuration-related code/unittests updated to support multiple configurations and the latest changes in mg-scripts.
There is a single test (test_replicates) that needs some additional updates.


qp_klp/tests/data/configuration_profiles/default.json
```diff
@@ -0,0 +1,91 @@
+{
```
antgonza, line 1: Why do we need this and the other json here? I thought they could be taken from mg-scripts  ... no?

qp_klp/tests/data/process_all_fastq_files.sh
```diff
@@ -0,0 +1,265 @@
+#!/bin/bash -l
```
antgonza, line 1: Why is this file now needed in the tests vs. just leaving it in the templates?

qp_klp/tests/data/configuration_profiles/default.json
charles-cowart: no. these are to support the unittests, which require initializing a pipeline object and hence a sample configuration file and configuration_profiles directory with sample profiles.

qp_klp/tests/data/process_all_fastq_files.sh
charles-cowart: it's _much_ less time-consuming to simply load the expected value from a file and compare it with the observed results rather than keep trying to make point changes to a list of strings that are often split to placate flake8. We really do modify that batch script often, at least currently.

antgonza: however, I thought we were not adding tests that compared this file in a tests ... no?

charles-cowart:
PR is ostensibly done. Reference to mg-scripts needs to point back to master once the latest mg-scripts PR has been reviewed, approved, and merged. Latest push didn't trigger CI, but includes the fixes to resolve all outstanding issues observed in CI.

qp_klp/tests/data/process_all_fastq_files.sh
charles-cowart: we're not adding new tests but this was an existing test. I could replace the test for one that simply checks for the existence of a job-script in the output and maybe checks the first few lines, but that seems a step backwards.

PR #81, opened 2024-02-27
bug-fixes after testing on qiita-rc

charles-cowart:
Here is a PR for the bug-fixes encountered after testing. There is one inconsequential change in mg-scripts for debugging purposes on qiita-rc so I did not include it in a second PR.

seqpro has so far taken four hours to count all of the files in our test job and it is still not finished. My assumption is that if seqpro completes successfully the steps downstream will be successful as well, as they're relatively small and were not affected by any recent changes.

seqpro keeps the generated sequence count for each file in memory. If the interpreter process dies, we can begin the Step() over again but all of the previous work will be lost. One possible solution might be to alter the demux() function to keep a count and write the info out to a json file that can later be read by a modified version of seqpro. This is similar to what it already does with other metadata files found in the run-directory. For existing jobs where NuQCJob has already run, we can write a job script to process them in parallel. That's where my thinking takes me.

### DIFF
--- a/qp_klp/Step.py
+++ b/qp_klp/Step.py
@@ -916,7 +916,7 @@ class Step:
             self.convert_bcl_to_fastq()
 
         increment_status()
-        if "QCJob" not in skip_steps:
+        if "NuQCJob" not in skip_steps:
             self.quality_control()
 
         increment_status()
--- a/qp_klp/klp.py
+++ b/qp_klp/klp.py
@@ -79,7 +79,8 @@ def sequence_processing_pipeline(qclient, job_id, parameters, out_dir):
     # successful.
 
     # at minimum, ConvertJob needs to have been successful.
-    is_restart = True if exists(join(out_dir, 'NuQCJob')) else False
+    out_path = partial(join, out_dir)
+    is_restart = True if exists(out_path('NuQCJob')) else False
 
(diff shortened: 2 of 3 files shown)


charles-cowart:
Regarding slow counting speed, it appears scikit-bio is known to be slow counting sequences, because sequences can still potentially be more or less than four lines (https://stackoverflow.com/questions/39150965/fastest-way-to-read-a-fastq-with-scikit-bio).

Selecting a 1.2GB file from the dataset, decompressing it, and getting a line count produced a result of approximately 17M sequences. The time taken by a small script using scikit-bio to pull the first 100,000 sequences from the zipped version of this file was a little over 44 seconds. If my back of the envelope estimate is correct counting all 162GB of data will take approximately 280 hours.

wasade:
Why not use `seqtk size` to count?

charles-cowart:
> seqtk size

Ty for the tip! Results are 17088755\t2516404944, the first component appears to match the numbers I got from wc -l /4. The results arrived in a little over 37 seconds:
real	0m37.218s
user	0m36.708s
sys	0m0.507s

charles-cowart:
I will need to put together a change to use seqtk in mg-scripts in a PR in the morning.

PR #84, opened 2024-03-27
failed-samples record is now recoverable.

charles-cowart:
FailedSamplesRecord() class now writes state out to file after each update. When an SPP run is manually restarted, failed_samples.html will now remain accurate.

For practical purposes, a small change made to the codebase during testing is also included here, as well as updates to tests due to recent changes in dependencies.


charles-cowart:
Resolves https://github.com/biocore/mg-scripts/issues/138

PR #86, opened 2024-04-02
Conditional fastq file finder fix

charles-cowart:
NuQC uses a helper function that must filter for non-fastq, zero-length, and other undesirable files. Depending on the type of pipeline, undesirables can also include index fastq files.
Fix allows I2 files to pass filter if they are found. New test tests for successful pass-through of I1 and I2 files when appropriate, as well as exclusion of zero-length, adapter-trimmed only files, etc.

### DIFF
--- a/qp_klp/tests/test_step.py
+++ b/qp_klp/tests/test_step.py
@@ -313,8 +313,14 @@ class BaseStepTests(TestCase):
         self.delete_these.append(tmp)
         return tmp
 
+    def _create_fake_file(self, path):
+        with open(path, 'w') as f:
+            f.write("This is a file.")
+
     def _create_test_input(self, stage):
         if stage >= 1:
+            # create an empty ConvertJob directory to test initialization
+            # with. Create fake binaries to test job submission.
             fake_path = join(self.output_file_path, 'ConvertJob', 'logs')
             makedirs(fake_path, exist_ok=True)
             fake_path = join(self.output_file_path, 'ConvertJob', 'Reports')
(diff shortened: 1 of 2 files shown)


qp_klp/tests/test_step.py
```diff
@@ -419,6 +426,68 @@ def _create_test_input(self, stage):
...
                 f.write("This is a file")
 
+        if stage >= 5:
+            for project in exp:
+                # currently unused for testing but pre-defined here in case
+                # the need arises.
+                # html_log_path = join(self.output_file_path, 'NuQCJob',
+                #                      project, 'fastp_reports_dir', 'html')
+                # json_log_path = join(self.output_file_path, 'NuQCJob',
+                #                      'project', 'fastp_reports_dir', 'json')
```
antgonza, line 436: rm these lines?

qp_klp/tests/test_step.py
```diff
@@ -419,6 +426,68 @@ def _create_test_input(self, stage):
             with open(fake_path, 'w') as f:
                 f.write("This is a file")
 
+        if stage >= 5:
```
antgonza, line 429: why 5?

antgonza (changes requested):

qp_klp/tests/test_step.py
charles-cowart: The 5 was so that it wouldn't interfere with legacy tests. Looking at it again there's something I want to tweak. I'll keep you posted.

PR #91, opened 2024-09-11
WIP: Added Qiita study reserved-word check

charles-cowart:
Depends on an additional PR to mg-scripts before passing tests.

### DIFF
--- a/qp_klp/tests/test_step.py
+++ b/qp_klp/tests/test_step.py
@@ -977,7 +977,7 @@ class BasicStepTests(BaseStepTests):
     def test_precheck(self):
         fake_client = AnotherFakeClient()
 
-        # test that precheck() raises a PipelineError with the correct
+        # test that Step.precheck() raises a PipelineError with the correct
         # message, given the configuration of Step() and AnotherFakeClient().
 
         step = Step(self.another_pipeline, self.qiita_id, None)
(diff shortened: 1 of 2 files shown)


qp_klp/tests/test_step.py
```diff
@@ -991,6 +991,24 @@ def test_precheck(self):
...
+
+        # _project_metadata_check() should return w/out raising an Error if
+        # step and fake_client is used.
+        step._project_metadata_check(fake_client)
+
+        fake_client.info_in_11661['categories'].append('well_id_384')
+        fake_client.info_in_13059['categories'].append('well_id_384')
+
+        msg = ("'well_id_384' exists in Qiita study 13059's sample metadata"
+               "\n'well_id_384' exists in Qiita study 11661's sample metadata")
```
antgonza, line 1008: My concern with this is that we are not actually testing the connection to qiita as you are using the fake_client, would it be possible to create a real one? Note that all plugins use this. See: https://github.com/qiita-spots/qp-woltka/blob/main/qp_woltka/tests/test_woltka.py#L26

antgonza (changes requested):
One comment.

qp_klp/tests/test_step.py
antgonza: After discussion offline, @charles-cowart confirmed that he's pointing to the right endpoint cause he tested outside of the tests + we agreed that reintroducing actual testing against the local/dev version will be in a different PR.

antgonza (approved):

PR #92, opened 2024-09-16
Refactored SPP plugin w/tellread support.

charles-cowart:
Refactored SPP plugin to support TellSeq. Replaced Step() workflow class and its two children (Metagenomic, Amplicon) w/Mixins. The new base Workflow() class is mixed with an Assay class and a SequencingTech class to create a new subclass e.g. StandardMetagenomicWorkflow(). Metatranscriptomic was also implemented as a separate class as well. All workflows are now much more extensible.

Job submissions and tests also updated to support sacct -> squeue migration and to support multiple Slurm jobs in parallel and wait on 1+ jobs until they complete or terminate.


qp_klp/StandardMetagenomicWorkflow.py
```diff
@@ -0,0 +1,184 @@
...
+            self.pre_check()
+
+        # this is performed even in the event of a restart.
+        self.generate_special_map()
+
+        # even if a job is being skipped, it's being skipped because it was
+        # determined that it already completed successfully. Hence,
+        # increment the status because we are still iterating through them.
+
+        self.update_status("Converting data", 1, 9)
```
antgonza, line 87: Should the 1 here be a counter and 9 be a variable so we updates in the future are easier?

qp_klp/tests/data/configuration_profiles/iseq_metagenomic.json
```diff
@@ -3,6 +3,25 @@
     "instrument_type": "iseq",
     "assay_type": "Metagenomic",
     "configuration": {
+      "tell-seq": {
+        "integrate_cores": "1",
+        "integrate_mem_limit": "8",
+        "integrate_script_path": "/home/qiita_test/qiita-spots/ADDITIONAL_REQUIREMENTS/integrate-indices-np.py",
```
antgonza, line 9: Would this be an issue for other testing? I mean, like testing in a laptop or in barnacle?

qp_klp/tests/data/configuration_profiles/miseq_metagenomic.json
```diff
@@ -64,7 +84,7 @@
         ],
         "fastqc_executable_path": "fastqc",
         "multiqc_executable_path": "multiqc",
-        "multiqc_config_file_path": "sequence_processing_pipeline/multiqc-bclconvert-config.yaml",
+        "multiqc_config_file_path": "/home/user/qp_klp/tests/data/multiqc-bclconvert-config.yaml",
```
antgonza, line 87: This should actually be an issue, no?

qp_klp/tests/data/sample-sheets/.DS_Store
antgonza, line 1: rm

qp_klp/tests/data/sample-sheets/metagenomic/.DS_Store
antgonza, line 1: rm

antgonza (changes requested):
A few questions.

qp_klp/Assays.py
```diff
@@ -0,0 +1,504 @@
...
+            project_name = project['project_name']
+            qiita_id = str(project['qiita_id'])
+
+            if qiita_id not in self.tube_id_map:
+                continue
+
+            # prep files are named in the following form:
+            # 20220423_FS10001773_12_BRB11603-0615.Matrix_Tube_LBM_14332.1.tsv
+            fqp_name = "%s_%s" % (project_name, qiita_id)
+            matching_files = [prep_file for prep_file in prep_file_paths if
```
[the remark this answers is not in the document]
charles-cowart: yes, because it's seqpro (metapool's command-line tool) that's responsible for namoing them.

qp_klp/Assays.py
```diff
@@ -0,0 +1,504 @@
...
+                prep_id = reply['prep']
+                results[study_id].append((prep_id, afact_name, is_repl))
+                self.run_prefixes[prep_id] = [metadata[sample]['run_prefix']
+                                              for sample in metadata]
+
+        self.touched_studies_prep_info = results
+        return results
+
+    def load_preps_into_qiita(self):
+        atype = 'FASTQ'
```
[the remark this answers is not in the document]
charles-cowart: Sure, that sounds good.

qp_klp/CHECKLIST.txt
```diff
@@ -0,0 +1,9 @@
+Confirm sif behavior works w/new KATHAROseq controls.
```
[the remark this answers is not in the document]
charles-cowart: Intentionally added. A todo list stored in the PR to make sure I wouldn't lose it. Since we squash before merge I figured it was okay to do this as long as I removed it beforehand.

qp_klp/StandardAmpliconWorkflow.py
```diff
@@ -0,0 +1,186 @@
...
+        self.is_restart = bool(self.kwargs['is_restart'])
+
+        if self.is_restart is True:
+            self.determine_steps_to_skip()
+
+        # this is a convenience member to allow testing w/out updating Qiita.
+        self.update = True
+
+        if 'update_qiita' in kwargs:
+            if bool(kwargs['update_qiita']) is False:
```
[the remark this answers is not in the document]
charles-cowart: Ty for the catch! Yes I should rework that for all of the Workflow versions.

qp_klp/StandardMetagenomicWorkflow.py
charles-cowart: We had that at one point but it meant defining the sequence of string messages in a separate location. It became much more difficult and error-prone to update them after adding or changing a step. Now it's clearer and pretty easy to search on update_status in this file and make adjustments that way.

qp_klp/tests/data/configuration_profiles/iseq_metagenomic.json
charles-cowart: Nope. It's just meant to reflect what an actual value might look like. it's not actually used in a test. In fact it would be better if I made it less specific.

qp_klp/SequencingTech.py
```diff
@@ -0,0 +1,256 @@
...
+        if hasattr(self, 'fsr'):
+            # NB 16S does not require a failed samples report and
+            # it is not performed by SPP.
+            self.fsr.write(failed_samples, job.__class__.__name__)
+
+        return failed_samples
+
+
+class TellSeq(SequencingTech):
+    seqtech_type = SEQTECH_NAME_TELLSEQ
```
wasade, line 91: I'm not sure I understand the class organization. Remember that TellSeq is a protocol which uses Illumina instruments

qp_klp/SequencingTech.py
```diff
@@ -0,0 +1,256 @@
...
+                              self.qiita_job_id)
+
+        job.run(callback=self.status_update_callback)
+
+        # audit the results to determine which samples failed to convert
+        # properly. Append these to the failed-samples report and also
+        # return the list directly to the caller.
+        failed_samples = job.audit(self.pipeline.get_sample_ids())
+        if hasattr(self, 'fsr'):
+            # NB 16S does not require a failed samples report and
```
wasade, line 146: We have not considered use of TellSeq for 16S

qp_klp/tests/data/sample-sheets/metagenomic/tellseq/good_sheet_absq_draft1.csv
```diff
@@ -0,0 +1,124 @@
...
+T_LS_7_8_15A,T.LS.7.8.15A,LS_Time_Series_ABSQ_P4,D3,TELLSeq_March_MetaG_Run_55555,C542,1,0.3750915,1.500366,168.9601351,LS_Time_Series_ABSQ_P4.T.LS.7.8.15A.D3
+,,,,,,,,,,
+[Bioinformatics],,,,,,,,,,
+Sample_Project,QiitaID,BarcodesAreRC,ForwardAdapter,ReverseAdapter,HumanFiltering,library_construction_protocol,experiment_design_description,contains_replicates,,
+TELLSeq_March_MetaG_Run_55555,55555,TRUE,GATCGGAAGAGCACACGTCTGAACTCCAGTCAC,GATCGGAAGAGCGTCGTGTAGGGAAAGGAGTGT,TRUE,Knight Lab Kapa HyperPlus,tellseq metagenomics,FALSE,,
+LS_Timeseries_TellSeq_10283,10283,TRUE,GATCGGAAGAGCACACGTCTGAACTCCAGTCAC,GATCGGAAGAGCGTCGTGTAGGGAAAGGAGTGT,TRUE,Knight Lab Kapa HyperPlus,tellseq metagenomics,FALSE,,
+,,,,,,,,,,
+[Contact],,,,,,,,,,
+Sample_Project,Email,,,,,,,,,
+TELLSeq_March_MetaG_Run_55555,cbrenchy@gmail.com,,,,,,,,,
```
wasade, line 119: There is a lot of information in here which I'm not sure is appropriate for a public repository

qp_klp/SequencingTech.py
charles-cowart: What I have is an observed need to be able to mix in either the legacy method to convert bcl files to fastq or the tellseq-based method to do the same; that's all that the classes are currently doing. I'm happy to go with whatever naming choices you suggest. Perhaps SequencingTech should be renamed to Protocol? Then TellSeq can keep its name and Illumina is renamed to Standard?

qp_klp/SequencingTech.py
charles-cowart: Gotcha, I'll remove the note from this class.

qp_klp/tests/data/configuration_profiles/miseq_metagenomic.json
charles-cowart: Not sure I understand -  it's just a dummy path.

qp_klp/tests/data/sample-sheets/metagenomic/tellseq/good_sheet_absq_draft1.csv
charles-cowart: Gotcha, I'll update it.

qp_klp/CHECKLIST.txt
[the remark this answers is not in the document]
antgonza: Got it, can you remove it?

qp_klp/tests/data/configuration_profiles/miseq_metagenomic.json
charles-cowart: correct.

charles-cowart:
@antgonza ty for your patience! It's ready for review!

antgonza (changes requested):
Some questions.

qp_klp/StandardAmpliconWorkflow.py
```diff
@@ -0,0 +1,189 @@
...
+        if 'update_qiita' in kwargs:
+            if not isinstance(kwargs['update_qiita'], bool):
+                raise ValueError("value for 'update_qiita' must be of "
+                                 "type bool")
+
+            self.update = kwargs['update_qiita']
+
+        # for now, hardcode this at the legacy value, since we've never
+        # changed it.
+        self.job_pool_size = 30
```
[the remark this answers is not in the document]
charles-cowart: It appears to be a value we set early on and never saw fit to change, at least for the elements that are using it. I put it back in to make the pipeline work in testing. I'll refactor it out.

qp_klp/klp.py
```diff
@@ -72,25 +72,35 @@ def sequence_processing_pipeline(qclient, job_id, parameters, out_dir):
...
+
+    # NB: RESTART FUNCTIONALITY:
+    # To restart a job that's failed, simply create a ProcessingJob() object
+    # in Qiita's interpreter using the Qiita Job ID aka the name of the
+    # working directory. _set_status("in construction") and then call
+    # submit(). New SPP jobs will record all the information needed to perform
+    # a restart should it fail.
     out_path = partial(join, out_dir)
-    is_restart = True if exists(out_path('NuQCJob')) else False
+    is_restart = True if exists(out_path('restart_me')) else False
```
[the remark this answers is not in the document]
charles-cowart: It's a good change actually. Since the user isn't supplying the three required parameters for restart and we previously weren't saving them, we had to scour the files in the working directory for them. This worked when we could extract the run-id from ConvertJob's sbatch file. This doesn't work for TellRead however.

Now every new job submission saves its parameters in 'restart_me' and if the job is restarted, klp can definitively say this is a restart or not.

qp_klp/tests/data/tellread_test.sbatch
```diff
@@ -0,0 +1,47 @@
...
+    -i qp_klp/tests/data/211021_A00000_0000_SAMPLE \
+    -o /home/runner/qp-knight-lab-processing/qp_klp/tests/data/077c4da8-74eb-4184-8860-0207f53623be/TellReadJob \
+    -s $(echo C501,C509,C502,C510,C503,C511,C504,C512,C505,C513,C506,C514,C507,C515,C508,C516,C517,C525,C518,C526,C519,C527,C520,C528,C521,C529,C522,C530,C523,C531,C524,C532,C533,C541,C534,C542,C535,C543,C536,C544,C537,C545,C538,C546,C539,C547,C540,C548,C549,C557,C550,C558,C551,C559,C552,C560,C553,C561,C554,C562,C555,C563,C556,C564,C565,C573,C566,C574,C567,C575,C568,C576,C569,C577,C570,C578,C571,C579,C572,C580,C581,C589,C582,C590,C583,C591,C584,C592,C585,C593,C586,C594,C587,C595,C588,C596 | tr -d '"') \
+    -g $(echo NONE,NONE,NONE,NONE,NONE,NONE,NONE,NONE,NONE,NONE,NONE,NONE,NONE,NONE,NONE,NONE,NONE,NONE,NONE,NONE,NONE,NONE,NONE,NONE,NONE,NONE,NONE,NONE,NONE,NONE,NONE,NONE,NONE,NONE,NONE,NONE,NONE,NONE,NONE,NONE,NONE,NONE,NONE,NONE,NONE,NONE,NONE,NONE,NONE,NONE,NONE,NONE,NONE,NONE,NONE,NONE,NONE,NONE,NONE,NONE,NONE,NONE,NONE,NONE,NONE,NONE,NONE,NONE,NONE,NONE,NONE,NONE,NONE,NONE,NONE,NONE,NONE,NONE,NONE,NONE,NONE,NONE,NONE,NONE,NONE,NONE,NONE,NONE,NONE,NONE,NONE,NONE,NONE,NONE,NONE,NONE | tr -d '"') \
```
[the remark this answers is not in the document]
charles-cowart: It will be possible when we add assemblies, possibly before.

qp_klp/__init__.py
```diff
@@ -13,7 +13,7 @@ class QiitaPluginAdmin(QiitaPlugin):
     _plugin_type = "private"
 
 
-__version__ = '2023.05'
+__version__ = '2024.11'
```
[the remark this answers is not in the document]
charles-cowart: Not sure I follow - the version number has been unchanged through several of the last releases. I updated it to 2024.11 for the upcoming release. However since it's now 2024.12, I can update it again. Does that sound all right?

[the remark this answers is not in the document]
charles-cowart: Release https://github.com/qiita-spots/qp-knight-lab-processing/releases/tag/2024.09 cut for previous release.

qp_klp/Assays.py
```diff
@@ -0,0 +1,511 @@
...
+        df = pd.DataFrame(data)
+        opath = join(self.pipeline.output_path, 'touched_studies.html')
+        with open(opath, 'w') as f:
+            f.write(df.to_html(border=2, index=False, justify="left",
+                               render_links=True, escape=False))
+
+        return df
+
+
+class MetaOmic(Assay):
```
antgonza, line 341: I'm not 100% why we need Assay and MetaOmic; could you clarify?

qp_klp/Assays.py
```diff
@@ -0,0 +1,511 @@
...
+    """
+    MetaOmic() is a base class for Metagenomic() and Metatranscriptomic(),
+    which are currently identical in functionality.
+    """
+    assay_type = ASSAY_NAME_NONE
```
antgonza, line 346: This is a duplicate as Assay should already have this and might lead to confusions, no?

qp_klp/Assays.py
```diff
@@ -0,0 +1,511 @@
...
+
+
+class Metagenomic(MetaOmic):
+    METAGENOMIC_TYPE = 'Metagenomic'
+    assay_type = ASSAY_NAME_METAGENOMIC
+
+
+class Metatranscriptomic(MetaOmic):
+    METATRANSCRIPTOMIC_TYPE = 'Metatranscriptomic'
+
```
antgonza, line 510: rm extra line

qp_klp/Protocol.py
```diff
@@ -0,0 +1,295 @@
...
+from sequence_processing_pipeline.SeqCountsJob import SeqCountsJob
+from sequence_processing_pipeline.TRIntegrateJob import TRIntegrateJob
+from sequence_processing_pipeline.PipelineError import PipelineError
+from os.path import join, split
+from re import match
+from os import makedirs, rename, walk
+from metapool import load_sample_sheet
+
+
+PROTOCOL_NAME_NONE = "None"
```
antgonza, line 12: This is never used again, except in line 24; why not just define there?

antgonza (changes requested):
A few more comments.

qp_klp/Assays.py
charles-cowart: Sure. Historically what we've had are two (Amplicon, Metagenomic) then three (Metatranscriptomic) pipelines that are largely similar with but with several large and small differences along the path. One example is Amplicon skipping the host-filtering step and then having to copy and reorganize files so that the output can be fed into FastQC and load preps into Qiita downstream. Metagenomic and Metatranscriptomic are essentially identical but they are required to have individual types. Now TellSeq is essentially the same pipeline downstream of host-filtering but the conversion process is entirely different and the host filtering needed to be modified to support its metadata.

Since we like to stay 'dry' and not repeat ourselves in code, we can't define four different pipelines that are 80% identical. Previous versions of SPP resolved this with an ever growing number of 'if type == 'Amplicon' else:' type statements. However introducing new features potentially broke all pathways and it became harder to understand how to modify the code to add new features or fix bugs.

What Assays does is break out all the code that changes depending on the pipeline being Amplicon or Meta*omic and organizes it so that it's all easier to read, changes to one type won't affect the other so it doesn't need to be retested, and we don't need to constantly test whether we're an amplicon run or not. Moreover helper methods that are shared between the different pipelines are defined in the base Assay class and don't have to be duplicated in each class.

qp_klp/Assays.py
charles-cowart: Good catch, thank you!

qp_klp/Protocol.py
charles-cowart: I'm taking a page out of metapool in that I'm starting to define global string constants at the top of the appropriate file. I actually have 'ASSAY_NAME_NONE' defined for Assays as well. It's true that it's currently only used in one case, but IMHO it's good to define it with the others so that they're all defined in the same place rather than, 'two cases up here and the one negative case down there where you're likely to forget about it.'

PR #93, opened 2024-10-18
Updates to bring plugin in-line w/mg-scripts

charles-cowart:


.github/workflows/qiita-plugin-ci.yml
```diff
@@ -79,7 +79,7 @@ jobs:
           # seqtk is a conda-installed requirement for metapool and isn't
           # installed automatically by its setup.py.
           conda config --add channels bioconda
-          conda create -q --yes -n klp python=3.9 seqtk
+          conda create -q --yes -n klp python=3.9 'seqtk>=1.4'
```
antgonza, line 82: should we open an issue to make sure we can use more recent versions?

qp_klp/Step.py
```diff
@@ -327,12 +327,13 @@ def _quality_control(self, config, input_file_path):
                            self.master_qiita_job_id,
                            config['job_max_array_length'],
                            config['known_adapters_path'],
+                           config['movi_executable_path'],
```
antgonza, line 330: just out of curiosity, why this change?

qp_klp/tests/data/process_all_fastq_files.sh
```diff
@@ -118,8 +118,8 @@ function mux-runner () {
 
     # minimap/samtools pair commands are now generated in NuQCJob._generate_mmi_filter_cmds()
     # and passed to this template.
-    minimap2 -2 -ax sr -t 1 /databases/minimap2/db_1.mmi ${jobd}/seqs.interleaved.fastq -a | samtools fastq -@ 1 -f 12 -F 256 > ${jobd}/foo
-minimap2 -2 -ax sr -t 1 /databases/minimap2/db_2.mmi ${jobd}/foo -a | samtools fastq -@ 1 -f 12 -F 256 > ${jobd}/bar
+    minimap2 -2 -ax sr -y -t 1 /databases/minimap2/db_1.mmi ${jobd}/seqs.interleaved.fastq -a | samtools fastq -@ 1 -f 12 -F 256 -T BX > ${jobd}/foo
```
antgonza, line 121: I think we should have a test without additional_fastq_tags; could you add?

antgonza (changes requested):
a few comments.

.github/workflows/qiita-plugin-ci.yml
charles-cowart: It should be fine. We only use seqtk to count sequences and I don't believe that would change from version to version. We're explicitly adding >= version 1.4 here because earlier versions don't have that functionality.

antgonza: Thank you for the clarification, I read that the wrong way (like 'seqtk<=1.4').

qp_klp/Step.py
charles-cowart: This goes back to a discussion we had with Amanda on the mg-scripts PR associated with this one. The group felt that default constructor parameters were not very desirable. I generally agree with that notion but I liked adding new parameters with defaults in order to preserve legacy behavior. Since I knew this was the sentiment I used the opportunity of adding a new parameter to go back and make the others positional as well.

qp_klp/tests/data/process_all_fastq_files.sh
charles-cowart: Sounds reasonable.

charles-cowart:
@antgonza We can change the reference to master mg-scripts after the other PR is approved and merged.

antgonza:
Superseded by https://github.com/qiita-spots/qp-knight-lab-processing/pull/98.

PR #96, opened 2024-12-13
fixes issue encountered during testing.

charles-cowart:


qp_klp/Protocol.py
```diff
@@ -129,33 +131,39 @@ def convert_raw_to_fastq(self):
...
+        files_to_count_path = join(self.pipeline.output_path,
+                                   'files_to_count.txt')
+
+        with open(files_to_count_path, 'w') as f:
+            # for raw_counts_r1r2, count corrected.err_barcode_removed files
+            # (TellReadJob final output).
+            for root, dirs, files in walk(self.raw_fastq_files_path):
+                for _file in files:
+                    if 'corrected.err_barcode_removed' in _file:
+                        print(join(root, _file), file=f)
```
antgonza, line 145: I don't really understand this step, it's just used for printing and not nothing gets written in the file, is that what you want to do?

antgonza (changes requested):
Just one question.

qp_klp/Protocol.py
```diff
@@ -129,33 +131,39 @@ def convert_raw_to_fastq(self):
...
+
+        files_to_count_path = join(self.pipeline.output_path,
+                                   'files_to_count.txt')
+
+        with open(files_to_count_path, 'w') as f:
+            # for raw_counts_r1r2, count corrected.err_barcode_removed files
+            # (TellReadJob final output).
+            for root, dirs, files in walk(self.raw_fastq_files_path):
+                for _file in files:
+                    if 'corrected.err_barcode_removed' in _file:
```
wasade, line 144: Why are these files being examined?

qp_klp/Protocol.py
charles-cowart: Actually, this is equivalent to writing f.write("%s\n" % join(root, _file)). There is an additional parameter "file=f" that redirects the print statement from stdout to the file handle 'f'. I've been using this form lately instead of constructing a string just for the sake of appending a newline.

qp_klp/Protocol.py
charles-cowart: yup, I have this marked in my notes. there is a patch for this that expects the integrated results directory instead of Full.

antgonza (approved):

PR #98, opened 2025-01-30
production hotfixes

charles-cowart:
Production hotfixes. Tests not yet updated.


qp_klp/Protocol.py
```diff
@@ -157,6 +165,11 @@ def generate_sequence_counts(self):
         if 'SeqCountsJob' not in self.skip_steps:
             job.run(callback=self.job_callback)
 
+        # if successful, set self.reports_path
+        self.reports_path = join(self.pipeline.output_path,
+                                 'SeqCountsJob',
+                                 'SeqCounts.csv')
```
wasade, line 171: Is this file consistent in schema with `Demultiplex_Stats.csv` used in the other object for `reports_path`?

qp_klp/StandardMetatranscriptomicWorkflow.py
```diff
@@ -74,45 +74,25 @@ def execute_pipeline(self):
...
 
         self.update_status("Generating preps", 4, 9)
-        if "GenPrepFileJob" not in self.skip_steps:
-            # preps are currently associated with array mixin, but only
-            # because there are currently some slight differences in how
-            # FastQCJob gets instantiated(). This could get moved into a
-            # shared method, but probably still in Assay.
-            self.generate_prep_file()
+
+        self.generate_prep_file()
```
wasade, line 95: It seems like the stuff in this method is the same as what's in the `StandardMetagenomicWorkflow`, can `execute_pipeline` be pushed to a base class?

qp_klp/Protocol.py
charles-cowart: No Demultiplex_Stats.csv's schema is defined and maintained by Illumina and carries additional stats beyond sequence counts. SeqCounts.csv is defined and maintained by us and just records the total raw-read counts for forward and reverse and the lane number per sample.

Consistency in this case is not an issue since the consumer of the metadata is the metapool module, and it only uses Demultuplex_Stats.csv for its raw-read counts. The equivalent output file for the bcl2fastq is similarly different. Our new method is more accurate in that it counts both the forward and reverse reads while I believe in Demultuplex_Stats.csv we just double the forward count.

qp_klp/StandardMetatranscriptomicWorkflow.py
charles-cowart: I could push much of both into the MetaOmics base class off the top of my head. Those two child classes are very similar. Currently they differ only in the string constants we define, off the top of my head. More than happy to include that in my next push.

qp_klp/Protocol.py
wasade: Isn't it risky, and unexpected, for an attribute common across subclasses to have a different interpretation?

qp_klp/StandardMetatranscriptomicWorkflow.py
wasade: Seems like a good idea?

antgonza: Moved.

PR #100, opened 2025-03-06
fixes after deploy 03.25

antgonza:


qp_klp/Assays.py
```diff
@@ -540,6 +567,8 @@ def execute_pipeline(self):
                 # (run-id, project_qid, other) and cleave
                 # the qiita-id from the project_name.
                 qid = _file.split('.')[1].split('_')[-1]
+                if not qid.isnumeric():
```
wasade, line 570: Do we know the structure of the filename sufficiently to express this as a regex?

antgonza: Not sure about regex and if adding that will actually help. Anyway, for clarity, as the comment says the file name is: run-id, project_qid, other - for example: 20231009_FS10001773_73_BTR67718-2515.U19_ADRC_14736.1.tsv so we could do something like '{self.pipeline.run_id}.{project_qid}.{extra}' where we are looking for the digits after '_', in this case '14736'. Does this help?

wasade: Why not just store the result of `match` in a variable and test whether it is `None`?

wasade (approved):

antgonza:
Thank you for the review @wasade !

PR #101, opened 2025-03-17
add execute_pipeline to StandardMetatranscriptomicWorkflow

antgonza:
This change was already deployed in qiita to avoid processing delays. Here I'm adding those changes and adding to the current tests the actual execution of the pipeline.


antgonza:
Thank you for review @wasade. Merging as the latests changes were just for CI proper run.

PR #102, opened 2025-03-20
Prep/Artifact human filtering

antgonza:
This piggybacks from the current workflows so we can human filter existing prep's raw data in Qiita. This version should be ready for testing in qiita-rc, which I'll do when available.

### DIFF
--- a/qp_klp/klp.py
+++ b/qp_klp/klp.py
@@ -6,14 +6,15 @@
 # The full license is in the file LICENSE, distributed with this software.
 # -----------------------------------------------------------------------------
 from functools import partial
-from os import environ
+from os import environ, symlink
 from qiita_client import ArtifactInfo
 from os import makedirs
-from os.path import join, exists
+from os.path import join, exists, basename
 from datetime import datetime
 from sequence_processing_pipeline.PipelineError import PipelineError
 from metapool import load_sample_sheet, MetagenomicSampleSheetv90
 import sample_sheet
+import pandas as pd
 from .Workflows import WorkflowError
 from .WorkflowFactory import WorkflowFactory
 
@@ -189,6 +190,7 @@ def prep_NuQCJob(qclient, job_id, parameters, out_dir):
 
     prep_info = qclient.get(f'/qiita_db/prep_template/{pid}/')
     dt = prep_info['data_type']
+    sid = prep_info['study']
     if dt not in {'Metagenomic', 'Metatranscriptomic'}:
         return (False, [], f'Prep {pid} has a not valid data type: {dt}')
     aid = prep_info['artifact']
@@ -196,10 +198,13 @@ def prep_NuQCJob(qclient, job_id, parameters, out_dir):
         return (False, [], f'Prep {pid} has a not valid artifact: {aid}')
 
     files, pt = qclient.artifact_and_preparation_files(aid)
+    pt.set_index('sample_name', inplace=True)
 
     out_path = partial(join, out_dir)
     final_results_path = out_path('final_results')
     makedirs(final_results_path, exist_ok=True)
+
+    project_name = f'qiita-{pid}-{aid}_{sid}'
 
     sheet = MetagenomicSampleSheetv90()
     sheet.Header['IEMFileVersion'] = '4'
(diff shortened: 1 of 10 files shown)


qp_klp/klp.py
```diff
@@ -158,3 +161,138 @@ def sequence_processing_pipeline(qclient, job_id, parameters, out_dir):
...
+    sheet.Header['SheetVersion'] = '90'
+    sheet.Header['Investigator Name'] = 'Qiita'
+    sheet.Header['Experiment Name'] = project_name
+
+    sheet.Bioinformatics = pd.DataFrame(
+        columns=['Sample_Project', 'ForwardAdapter', 'ReverseAdapter',
+                 'library_construction_protocol',
+                 'experiment_design_description',
+                 'PolyGTrimming', 'HumanFiltering', 'QiitaID'],
+        data=[[project_name, 'NA', 'NA', 'NA', 'NA', 'FALSE', 'TRUE', sid]])
```
wasade, line 227: Is there a reason this stuff isn't done by the `MetagenomicSampleSheetv90` constructor?

qp_klp/klp.py
```diff
@@ -158,3 +161,138 @@ def sequence_processing_pipeline(qclient, job_id, parameters, out_dir):
...
+            'Sample_Plate': '',
+            'well_id_384': '',
+            'I7_Index_ID': '',
+            'index': vals['index'],
+            'I5_Index_ID': '',
+            'index2': vals['index2'],
+            'Sample_Project': project_name,
+            'Well_description': '',
+            'Sample_Well': '',
+            'Lane': '1'}
```
wasade, line 243: It seems like a lot of what's here are attributes specific to the sample sheet. Is there a reason this is done outside of the sample sheet, rather than controlled by the sample sheet?

qp_klp/klp.py
```diff
@@ -158,3 +161,138 @@ def sequence_processing_pipeline(qclient, job_id, parameters, out_dir):
...
+        sheet.add_sample(sample_sheet.Sample(sample))
+
+    sheet.Contact = pd.DataFrame(
+        columns=['Email', 'Sample_Project'],
+        data=[['qiita.help@gmail.com', project_name]])
```
wasade, line 248: Same here. Or this perhaps could be `sheet.set_contact(email, project_name)`

qp_klp/klp.py
```diff
@@ -158,3 +161,138 @@ def sequence_processing_pipeline(qclient, job_id, parameters, out_dir):
...
+    with open(new_sample_sheet, 'w') as f:
+        sheet.write(f, 1)
+
+    # now that we have a sample_sheet we can fake the
+    # ConvertJob folder so we are ready for the restart
+    convert_path = out_path('ConvertJob')
+    project_folder = out_path('ConvertJob', project_name)
+    makedirs(project_folder, exist_ok=True)
+
+    for _, fs in files.items():
```
wasade, line 260:
```suggestion
    for fs in files.values():
```

qp_klp/klp.py
```diff
@@ -158,3 +161,138 @@ def sequence_processing_pipeline(qclient, job_id, parameters, out_dir):
...
+        for f in fs:
+            bn = basename(f['filepath'])
+            symlink(f['filepath'], f'{project_folder}/{bn}')
+
+    # create job_completed file to skip this step
+    with open(f'{convert_path}/job_completed', 'w') as f:
+        f.write('')
```
wasade, line 267: See [SO](https://stackoverflow.com/a/34603829) post, but `Path(f'{convert_path}/job_completed').touch()`

qp_klp/klp.py
```diff
@@ -158,3 +161,138 @@ def sequence_processing_pipeline(qclient, job_id, parameters, out_dir):
...
+                  'lane_number': "1",
+                  'config_fp': CONFIG_FP,
+                  'run_identifier': '211021_A00000_0000_SAMPLE',
+                  'output_dir': out_dir,
+                  'job_id': job_id,
+                  'status_update_callback': status_line.update_job_status,
+                  # set 'update_qiita' to False to avoid updating Qiita DB
+                  # and copying files into uploads dir. Useful for testing.
+                  'update_qiita': True,
+                  'is_restart': True}
```
wasade, line 281:
```suggestion
    kwargs = {'qclient': qclient,
              'uif_path': new_sample_sheet,
              'lane_number': "1",
              'config_fp': CONFIG_FP,
              'run_identifier': '211021_A00000_0000_SAMPLE',
              'output_dir': out_dir,
              'job_id': job_id,
              'status_update_callback': status_line.update_job_status,
              # set 'update_qiita' to False to avoid updating Qiita DB
              # and copying files into uploads dir. Useful for testing.
              'update_qiita': True,
              'is_restart': True}
    try:
```

qp_klp/klp.py
```diff
@@ -158,3 +161,138 @@ def sequence_processing_pipeline(qclient, job_id, parameters, out_dir):
...
+        status_line.update_job_status("Getting project information")
+
+        workflow.execute_pipeline()
+
+        status_line.update_job_status("SPP finished")
+
+    except (PipelineError, WorkflowError) as e:
+        # assume AttributeErrors are issues w/bad sample-sheets or
+        # mapping-files.
+        return False, None, str(e)
```
wasade, line 294: The traceback is lost here:

```
In [4]: def foo():
   ...:     raise ValueError("stuff")
   ...: 

In [5]: try:
   ...:     foo()
   ...: except ValueError as e:
   ...:     print(str(e))
   ...: 
stuff
```

See this [SO](https://stackoverflow.com/a/3702847) post. I think the intent is `str(e)` to instead be `traceback.format_exc()`, right? If so then `import traceback` is also needed

qp_klp/tests/test_prep_nuqcjob.py
```diff
@@ -0,0 +1,107 @@
...
+        self.assertFalse(success)
+
+        # test success; note that success at this point is that
+        # the job fails when submitting via sbatch
+        pid, job_id = self._setup_test()
+        success, ainfo, msg = prep_NuQCJob(
+            self.qclient, job_id, {'prep_id': pid}, out_dir)
+        self.assertTrue(
+            msg.startswith("Execute command-line statement failure:"))
+        self.assertTrue('sbatch: command not found' in msg)
```
wasade, line 102: Could a test be included which asserts successful execution?

wasade (changes requested):
Thanks! It seems like some of the logic here would be better placed in the sample sheet object, and it would be valuable to have a test asserting successful execution

antgonza:
@wasade, thank you for the prompt review! 

I agree on moving the main parts of the code to a new object/class but note that the main difference is that the current objects read a sample-sheet, while the code creates a sample-sheet and then uses that to process the reads. 

Now, allowing for a full run in the github CI might be really difficult as it depends on slurm and running the full pipeline (which some steps are really compute intensive) - IMOO, we should tests correct execution on qiita-rc; from start to finish. The good news is that I already have this mostly written and I could add that "script" to this repo in a future PR.

wasade:
I wasn't suggesting a new object, but to place it in the sample sheet objects? I may not know enough about this project though to comment sufficiently, other than it looked like object specific responsibilities might be handled outside of the object.

`sbatch` can be thought of as fancy `bash`, why not just mock it out so the commands can run? If the concern is runtime in CI, then one solution would be to run a subset of the data.

An example of mocking out `sbatch` can be found [here](https://github.com/biocore/microsetta-processing/blob/main/scripts/submit_all.sh#L74-L78) with the capture [here](https://github.com/biocore/microsetta-processing/blob/main/scripts/sbatch-capture.sh). It assumes use of `sbatch` is by `PIPE` although a similar type of approach could be used for a `.sbatch` script if that's what the framework here does (but again, I don't know much about this project)

antgonza:
Thank you.

Agree on "other than it looked like object specific responsibilities"; I'll move the code around.

I'll also add an issue to "mock out sbatch for pipeline testing" so we don't forget about it but also do not stop this branch to move forward as I consider that outside the scope of this branch.

qp_klp/klp.py
antgonza: I just want to make sure I'm not missing something obvious but are you sure this method exists or are you asking why not put it together?

FWIW, neither the original [sample_sheet](https://github.com/clintval/sample-sheet) or the augmented version[ metapool](https://github.com/biocore/metagenomics_pooling_notebook/) provides it:
```python
In [1]: import sample_sheet

In [2]: sample_sheet.set_contact
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
Cell In[2], line 1
----> 1 sample_sheet.set_contact

AttributeError: module 'sample_sheet' has no attribute 'set_contact'

In [3]: from metapool import MetagenomicSampleSheetv90

In [4]: MetagenomicSampleSheetv90.set_contact
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
Cell In[4], line 1
----> 1 MetagenomicSampleSheetv90.set_contact

AttributeError: type object 'MetagenomicSampleSheetv90' has no attribute 'set_contact'
```

Now, just to be clear, the idea of this code is to quickly create a "valid" sample_sheet from a loaded prep in qiita so we can human-filter/qc their files.

wasade: If the method doesn't exist, then for the same scope of responsibility reasons noted elsewhere, creating it seems like it would improve the overall codebase. In general, I advise making these type of low risk easy changes (such as this and abstracting sbatch) under the principle of continuous improvement, rather than creating issues and deferring. But you are the lead on this project so it is your decision about how to proceed.

qp_klp/klp.py
antgonza: AFAIK, the intent is completely the oposite, just raise/report the error in the GUI with minimal information for the user. I think that way, if is something obvious and handled, like "sample sheet has wrong OverwriteCycles value", that's what's shown but if there is something less obvious, users will need to contact the admins/devs to investigate. FWIW, this has been a useful interaction for this specific plugin: wet/dry-lab interactions.

wasade: How do the dev's know what line in the codebase is throwing the exception without the traceback?

wasade: But the developer will not know the exact line of code raising the exception. Won't that require the developer to then guess or perform a much more time expensive debugging process to determine what specifically failed?

antgonza: I see what you are saying but in experience so far that's not the case. However, we might be missing something so I'll change and we can revert back if users get too annoyed.

wasade: thanks! For users, the traceback could either be post processed, or an additional item in the tuple could be returned (the original `str(e)`)

antgonza: Yeah, decided to write a log in the outdir so devs can see the full traceback but keep it simple ("str(e)") for users.

antgonza:
Closing in favor of: https://github.com/qiita-spots/qp-knight-lab-processing/pull/132

PR #129, opened 2025-04-22
inserting data to qiita for testing

antgonza:


AmandaBirmingham (changes requested):
Hi, @antgonza --the only concern I have her is with adding what appears (maybe?) to be real info as test data, particularly in the sql .  Are all the qiita studies and metadata that are added here already public in qiita?  If I'm reading Qiita right, it seems like at least one isn't, although I didn't check them all.  @wasade had raised a concern about this in the past in other repos, so I think we want to avoid repeating that problem in this one!

antgonza:
Thank you for your review. IMOO this is not a concern as the sample-sheet doesn't contain any actual metadata or identifying information. However, happy to change whatever you think might not be OK, just let me know.

src/qp_klp/Workflows.py
```diff
@@ -517,8 +517,9 @@ def _compare_samples_against_qiita(self):
                 # check for possible match w/tube-ids, if defined in project.
                 results_tid = self._process_tube_ids(qiita_id, samples)
                 if results_tid:
-                    msgs.append("Number of values in sheet that aren't "
-                                "tube-ids in Qiita: %s" % len(results_tid[0]))
+                    rtid = results_tid[0]
+                    msgs.append('Number of sample-names not in Qiita: '
```
AmandaBirmingham, line 521: I don't think this revised message is providing all the necessary info, which is that not only are these identifiers not found as sample names in qiita, they are also not found as tube ids in qiita.  I definitely would prefer that it not be identical in wording to the message at 507.  Can you modify?

AmandaBirmingham (changes requested):
One small request for extension to an error message. Thank you for all the test file changes--I am quite aware of what a pain they are!

src/qp_klp/Workflows.py
antgonza: Yes!

tests/test_helpers.py
```diff
@@ -402,3 +402,7 @@ def test_failed_samples_record(self):
         self.assertFalse(exists(fsr.report_path))
         fsr.generate_report()
         self.assertTrue(exists(fsr.report_path))
+
+
+if __name__ == '__main__':
+    main()
```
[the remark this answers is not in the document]
antgonza: I don't like to run all tests via nosetests locally, I prefer to just to python [file-to-test], for that to work you need those 2 lines.

AmandaBirmingham (approved):

PR #104, opened 2025-04-03
rm duplicated execute_pipeline

antgonza:

PR #131, opened 2025-05-01
test full runs in WorkflowFactory.py

antgonza:


tests/qiita-sql/91.sql
```diff
@@ -14,6 +14,19 @@ INSERT INTO
         study_abstract
     )
 VALUES
+    (
```
antgonza, line 17: Amplicon actually inserts preps to study 1 (exists as default in qiita) and 2, which doesn't exist.

antgonza:
@AmandaBirmingham, this is ready for review. This supersedes https://github.com/qiita-spots/qp-knight-lab-processing/pull/104. This version is actually running all the pipelines to the end: submitting new preps to Qiita; which is nice and should help for future improvements; this additionally means that now we know all the steps for all pipelines and all of them have the same steps (even if they are just "pass").

tests/test_WorkflowFactory.py
```diff
@@ -123,8 +216,14 @@ def test_metagenomic_workflow_creation(self):
         self.assertEqual(wf.protocol_type, PROTOCOL_NAME_ILLUMINA)
         self.assertEqual(wf.assay_type, ASSAY_NAME_METAGENOMIC)
 
-        with self.assertRaisesRegex(
-                PipelineError, 'There are no fastq files for FastQCJob'):
+        self._inject_data(wf)
+
+        # Metagenomic is a valid data type in the default qiita test
+        # database but job-id: 78901 doesn't exist; however, if we get
+        # to here, it means that all steps have ran to completion
```
AmandaBirmingham, line 223: typo: should be "have run"

tests/test_WorkflowFactory.py
```diff
@@ -147,15 +246,20 @@ def test_metatranscriptomic_workflow_creation(self):
         self.assertEqual(wf.protocol_type, PROTOCOL_NAME_ILLUMINA)
         self.assertEqual(wf.assay_type, ASSAY_NAME_METATRANSCRIPTOMIC)
 
-        with self.assertRaisesRegex(
-                PipelineError, 'There are no fastq files for FastQCJob'):
+        self._inject_data(wf)
+        # Metatranscriptomic is not a valid data type in the default qiita test
+        # database; however, if we get to here, it means that all steps have
+        # ran to completion and the system is trying to create the preps.
```
AmandaBirmingham, line 252: same typo as above

tests/test_WorkflowFactory.py
```diff
@@ -123,8 +216,14 @@ def test_metagenomic_workflow_creation(self):
...
-        with self.assertRaisesRegex(
-                PipelineError, 'There are no fastq files for FastQCJob'):
+        self._inject_data(wf)
+
+        # Metagenomic is a valid data type in the default qiita test
+        # database but job-id: 78901 doesn't exist; however, if we get
+        # to here, it means that all steps have ran to completion
+        # and the system is trying to create the preps.
+        with self.assertRaisesRegex(RuntimeError, 'invalid input '
+                                    'syntax for type uuid: "78901"'):
```
AmandaBirmingham, line 226: I am confused by the error message and the comment here.  What does "invalid input syntax for type uuid <whatever>" *mean*?  The comment mentions that 78901 is a job id, so why does the error message refer to it as a "type uuid"?

tests/test_WorkflowFactory.py
```diff
@@ -169,8 +273,47 @@ def test_amplicon_workflow_creation(self):
...
+        makedirs(fastqc_dir, exist_ok=True)
+        makedirs(multiqc_dir, exist_ok=True)
+        # generating the "*.completed" files
+        for i in range(3):
+            Path(f'{fastqc_dir}/FastQCJob_{i}.completed').touch()
+            Path(f'{multiqc_dir}/MultiQCJob_{i}.completed').touch()
+
+        # Amplicon is a valid data type in the default qiita test
+        # database but job-id: 78901 doesn't exist; however, if we get
+        # to here, it means that all steps have ran to completion
```
AmandaBirmingham, line 313: same typo as above

tests/test_WorkflowFactory.py
```diff
@@ -193,9 +336,14 @@ def test_tellseq_workflow_creation(self):
...
         self.assertEqual(wf.assay_type, ASSAY_NAME_METAGENOMIC)
 
-        with self.assertRaisesRegex(
-                FileNotFoundError,
-                'TellReadJob/sample_index_list_TellReadJob.txt'):
+        self._inject_data(wf)
+
+        # Metagenomic is a valid data type in the default qiita test
+        # database but job-id: 78901 doesn't exist; however, if we get
+        # to here, it means that all steps have ran to completion
```
AmandaBirmingham, line 343: same typo as above

src/qp_klp/Assays.py
```diff
@@ -144,6 +143,141 @@ def _generate_artifact_name(self, prep_file_path):
             # this is a normal pre-prep or sample-sheet.
             return (a_name, False)
 
+    def post_process_raw_fastq_output(self):
+        """This is an Amplicon specific method, see that object for
+           more info
+        """
+        pass
+
+    def quality_control(self):
```
AmandaBirmingham, line 152: So, `post_process_raw_fastq_output` is only for amplicons and `quality_control` is only for metaomics; they are mutually exclusive, and whichever one the assay in question has gets called in "qc-ing reads" step of `execute_pipeline`:
```
self.update_status("QC-ing reads", 2, 9)
if "NuQCJob" not in self.skip_steps:
            self.post_process_raw_fastq_output()
            self.quality_control()
```

Given this, why have two separate methods?  It seems like we could have a `def qc_reads()` method on `Assay` with no default contents and we could override it in `Amplicon` with whatever is currently in`Amplicon.post_process_raw_fastq_output` and in `MetaOmics` with whatever is currently in `MetaOmics.quality_control`.  Then `execute_pipeline` could just say:

```
self.update_status("QC-ing reads", 2, 9)
if "NuQCJob" not in self.skip_steps:
           self.qc_reads()
```

src/qp_klp/Assays.py
```diff
@@ -144,6 +143,141 @@ def _generate_artifact_name(self, prep_file_path):
...
+            # only because metagenomic runs currently require a failed-samples
+            # report to be generated. This is not done for amplicon runs since
+            # demultiplexing occurs downstream of SPP.
+            self.generate_reports()
+
+        self.update_status("Generating preps", 4, 9)
+        if "GenPrepFileJob" not in self.skip_steps:
+            # preps are currently associated with array mixin, but only
+            # because there are currently some slight differences in how
+            # FastQCJob gets instantiated(). This could get moved into a
```
AmandaBirmingham, line 200: A little confused by the mention of `FastQCJob` here since I thought we handled that at line 189 ... is this some sort of downstream consequence of that stuff?

src/qp_klp/Assays.py
```diff
@@ -144,6 +143,141 @@ def _generate_artifact_name(self, prep_file_path):
...
+        self.update_status("Loading preps into Qiita", 7, 9)
+        if self.update:
+            self.update_prep_templates()
+
+        # before we load preps into Qiita we need to copy the fastq
+        # files n times for n preps and correct the file-paths each
+        # prep is pointing to.
+        self.load_preps_into_qiita()
+
+        self.fsr.generate_report()
```
AmandaBirmingham, line 272: Could we toss this line a comment, too? I see that it was in `MetaOmic.execute_pipeline` but not in `StandardAmpliconWorkflow.execute_pipeline`, so now I'm curious about why ... :D. I also notice that at one point in `MetaOmic` the code checks `if hasattr(self, 'fsr'):` before calling `self.fsr.<something>()` ... do we need to worry that not every assay type or every individual assay instance will have an `fsr` property?

AmandaBirmingham (commented):
A few questions and a typo :)

tests/test_WorkflowFactory.py
antgonza: Qiita job ids are UUID in the database, this tests uses 78901 as the job id so the database complains about the format. IMOO it is fine to let it fail here as is the last step of the processing.

src/qp_klp/Assays.py
antgonza: removing, left behind by mistake.

src/qp_klp/Assays.py
antgonza: Great catch and suggestion, thank you!

src/qp_klp/Assays.py
antgonza: fsr is an instance of FailedSamplesRecord, which is its own object, and as far as I can tell is used with Job.audit to keep track of the samples lost on each of the steps of the pipeline. Here we call FailedSamplesRecord. generate_report to output the report so it's moved to the final output and the user can get access. Adding this info in the code.

tests/test_WorkflowFactory.py
AmandaBirmingham: Aha, got it!  For the benefit of non-experts like me reading the tests, would you consider having these repeated comments say essentially what you just did? 

```
        # Metagenomic is a valid data type in the default qiita test
        # database but the test job-id, 78901, isn't in the 
        # format of a uuid as expected; however, if we get
        # to here, it means that all steps have run to completion
```

AmandaBirmingham (commented):
Thank you for the changes and explanations.  One more request for changes--only to comments, though!--and it looks good to me.

AmandaBirmingham (approved):

antgonza:
Thank you @AmandaBirmingham !

PR #132, opened 2025-05-08
adding PrepNuQCWorkflow

antgonza:


src/qp_klp/Assays.py
```diff
@@ -148,8 +148,8 @@ def execute_pipeline(self):
         Executes steps of pipeline in proper sequence.
         :return: None
         '''
-        if not self.is_restart:
-            self.pre_check()
+        # pre_check-ing the status of the workflow
+        self.pre_check()
```
AmandaBirmingham, line 152: why is it no longer necessary to skip this on a restart?

src/qp_klp/StandardMetagenomicWorkflow.py
```diff
@@ -49,3 +60,141 @@ def __init__(self, **kwargs):
...
+        sheet.Header['SheetType'] = 'standard_metag'
+        sheet.Header['SheetVersion'] = '90'
+        sheet.Header['Investigator Name'] = 'Qiita'
+        sheet.Header['Experiment Name'] = project_name
+
+        sheet.Bioinformatics = pd.DataFrame(
+            columns=['Sample_Project', 'ForwardAdapter', 'ReverseAdapter',
+                     'library_construction_protocol',
+                     'experiment_design_description',
+                     'PolyGTrimming', 'HumanFiltering', 'QiitaID'],
```
AmandaBirmingham, line 117: `'PolyGTrimming'` is not one of the bioinformatics fields that any of the existing sample sheet types (including v90) know about ... this is what we get for not enforcing validation.  Can you educate me on what this field is for and what sort(s) of sample sheets use it?

src/qp_klp/StandardMetagenomicWorkflow.py
```diff
@@ -49,3 +60,141 @@ def __init__(self, **kwargs):
...
+        sheet.Contact = pd.DataFrame(
+            columns=['Email', 'Sample_Project'],
+            data=[['qiita.help@gmail.com', project_name]])
+
+        new_sample_sheet = out_path('sample-sheet.csv')
+        with open(new_sample_sheet, 'w') as f:
+            sheet.write(f, 1)
+
+        # now that we have a sample_sheet we can fake the
+        # ConvertJob folder so we are ready for the restart
```
AmandaBirmingham, line 167: the "so we are ready for the restart" comment here is opaque to me--is this class only used in cases with a restart? or is there only a restart sometimes?

src/qp_klp/klp.py
```diff
@@ -127,34 +129,70 @@ def sequence_processing_pipeline(qclient, job_id, parameters, out_dir):
...
-
         workflow.execute_pipeline()
-
-        status_line.update_job_status("SPP finished")
-
     except (PipelineError, WorkflowError) as e:
-        # assume AttributeErrors are issues w/bad sample-sheets or
-        # mapping-files.
+        # add full traceback to the working directory so devs/admins
+        # can review and get as much info as possible
```
AmandaBirmingham, line 150: +1

src/sequence_processing_pipeline/Commands.py
```diff
@@ -106,7 +106,13 @@ def demux(id_map, fp, out_d, task, maxtask):
             fullname_r1 = outdir + sep + r1 + ext
             fullname_r2 = outdir + sep + r2 + ext
 
-            os.makedirs(outdir, exist_ok=True)
+            # we have seen in lustre that sometime this line
+            # can have a raise condition; making sure it doesn't brake
```
AmandaBirmingham, line 110: typo: replace "brake" with "break"

src/sequence_processing_pipeline/Pipeline.py
```diff
@@ -267,8 +267,8 @@ def __init__(self, configuration_file_path, run_id, input_file_path,
         required_files = ['RTAComplete.txt', 'RunInfo.xml']
         for some_file in required_files:
             if not exists(join(self.run_dir, some_file)):
-                raise PipelineError("required file '%s' is not present." %
-                                    some_file)
+                raise PipelineError(f"required file '{some_file}' is not "
+                                    f"present {self.run_dir}.")
```
AmandaBirmingham, line 271: For readability, how about "is not present **in** {self.run_dir}."?

AmandaBirmingham (changes requested):
I have no idea what this new PrepNuQC stuff does, so I will take your word for it that it is doing it right :) I do, however, have strong opinions about a couple of implementation details that intersect with kl-metapool functionality.  Let me know if you want to discuss!

src/qp_klp/StandardMetagenomicWorkflow.py
```diff
@@ -49,3 +60,141 @@ def __init__(self, **kwargs):
...
+        files, pt = qclient.artifact_and_preparation_files(aid)
+        html_summary = qclient.get_artifact_html_summary(aid)
+        if html_summary is None:
+            raise WorkflowError(f'Artifact {aid} doesnot have a summary, '
+                                'please generate one.')
+        df_summary = pd.read_html(html_summary)[0]
+        pt.set_index('sample_name', inplace=True)
+
+        project_name = f'qiita-{pid}-{aid}_{sid}'
+        run_identifier = '250225_LH00444_0301_B22N7T2LT4'
```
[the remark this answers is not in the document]
antgonza: Because the code checks that a real run-id is used and as this is piggy backing on the existing Workflow to do the filtering, it uses something that already exists. FWIW, this is being added also in the tests. I'll add a comment in the code about this.

src/qp_klp/StandardMetagenomicWorkflow.py
```diff
@@ -49,3 +60,141 @@ def __init__(self, **kwargs):
...
+
+        sheet = MetagenomicSampleSheetv90()
```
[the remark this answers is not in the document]
antgonza: Sure, I don't think it will make a difference for this code. I'll update it and see if the test pass.

src/qp_klp/StandardMetagenomicWorkflow.py
```diff
@@ -49,3 +60,141 @@ def __init__(self, **kwargs):
...
+        sheet.Header['IEMFileVersion'] = '4'
+        sheet.Header['Date'] = datetime.today().strftime('%m/%d/%y')
+        sheet.Header['Workflow'] = 'GenerateFASTQ'
+        sheet.Header['Application'] = 'FASTQ Only'
+        sheet.Header['Assay'] = prep_info['data_type']
+        sheet.Header['Description'] = f'prep_NuQCJob - {pid}'
+        sheet.Header['Chemistry'] = 'Default'
+        sheet.Header['SheetType'] = 'standard_metag'
```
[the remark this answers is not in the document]
antgonza: Had no idea that method existed; thank you for the suggestion. I'll try to incorporate it.

src/qp_klp/StandardMetagenomicWorkflow.py
antgonza: Interesting, I guess I used an old version of a sample-sheet when I wrote this. Looking in my emails, I have an email saying that in a new sample-sheet I should  """Remove one column, it’s not used anymore: PolyGTrimming""". I can remove it.

src/qp_klp/StandardMetagenomicWorkflow.py
antgonza: Yes, it's always a restart as we already have the FASTQ files created and we don't need to start from scratch.

antgonza (commented):
@AmandaBirmingham, thank you for the review.

antgonza:
Fantastic, thank you!

PR #133, opened 2025-06-01
minor fixes 0525

antgonza:


src/qp_klp/Assays.py
```diff
@@ -590,6 +590,25 @@ def update_prep_templates(self):
             for prep_fp in self.prep_file_paths[study_id]:
                 metadata = Assay._parse_prep_file(prep_fp)
                 afact_name, is_repl = self._generate_artifact_name(prep_fp)
+
+                # the preps are created by seqpro within the GenPrepFileJob;
```
antgonza, line 594: FWIW, This is currently only used in PrepNuQCWorkflow

src/sequence_processing_pipeline/Pipeline.py
```diff
@@ -911,8 +911,8 @@ def is_sample_sheet(sample_sheet_path):
 
         return False
 
-    def _generate_dummy_sample_sheet(self, first_read, last_read,
-                                     indexed_reads, dummy_sample_id):
+    def _generate_dummy_sample_sheet(self, index_cycles, non_index_cycles,
```
antgonza, line 914: FWIW, the code expected (as this is how old RunInfo.xml files are formed) that the reads would be like index1, read1, read2, index2 - however this is not the case in the new versions. Thus, changed the method to actually be aware of what's an index and non_index read

antgonza (commented):
Just adding some comments for easier review.

src/qp_klp/WorkflowFactory.py
```diff
@@ -19,6 +19,7 @@ class WorkflowFactory():
     ST_TO_IN_MAP = {PROTOCOL_NAME_ILLUMINA: ['standard_metag',
                                              'standard_metat',
                                              'absquant_metag',
+                                             'abs_quant_metag',
```
AmandaBirmingham, line 22: Given that this module already depends on `metapool`, I would really, **really** like it if, instead of redefining strings for the same sheet-type names that are (canonically) defined in `metapool.sample_sheet`, we just used the public values defined there:

https://github.com/biocore/kl-metapool/blob/7bb932f1cb8a69e49c087710e14de046eaef0079/metapool/sample_sheet.py#L43-L47

Also, there still is no `absquant_metat` (and there are no plans that there ever will be).  Using (only) the sheet types from in the module where we actually define the sheet types would ensure that we never have this sort of nonsense entry in the future.

src/sequence_processing_pipeline/Pipeline.py
```diff
@@ -28,7 +28,7 @@
 class InstrumentUtils():
     types = {'A': 'NovaSeq 6000', 'D': 'HiSeq 2500', 'FS': 'iSeq',
              'K': 'HiSeq 4000', 'LH': 'NovaSeq X Plus', 'M': 'MiSeq',
-             'MN': 'MiniSeq',
+             'MN': 'MiniSeq', 'SH': 'MiSeq i100',
```
AmandaBirmingham, line 31: I am worried about this one, not for this code, but for `kl-metapool`, which AFAICT doesn't know about MiSeq i100 (as opposed to just regular MiSeq).  What do we know about whether it needs its i5 barcode revcomped or not, etc?  Is `kl-metapool` going to need modifications to handle this sequencer?

src/sequence_processing_pipeline/Pipeline.py
AmandaBirmingham: I am 100% in favor of anything that makes the nature of the inputs more explicit! However, I don't really understand what the inputs should be here. Is it possible to add a docstring for this function, maybe with an example input for each param?

src/sequence_processing_pipeline/Pipeline.py
```diff
@@ -973,32 +973,21 @@ def generate_dummy_sample_sheet(self, run_dir, output_fp):
...
-            # in the middle and maximum two in the middle.
-            raise ValueError("RunInfo.xml contains abnormal reads.")
-
-        first_read = reads.pop(0)
-        last_read = reads.pop()
-
-        if (first_read['IsIndexedRead'] is True or
-                last_read['IsIndexedRead'] is True):
+        # the assumptions are: if we have 3 reads we should only have 1
+        # index; and if we have 4 reads, 2 should be index
```
AmandaBirmingham, line 977: +1!

tests/data/RunInfo_Good3.xml
```diff
@@ -0,0 +1,26 @@
+<?xml version="1.0"?>
```
AmandaBirmingham, line 1: Is this one from a NovaSeqXPlus?

AmandaBirmingham (changes requested):
Small requests. Thanks!

src/qp_klp/Assays.py
```diff
@@ -478,6 +478,8 @@ def qc_reads(self):
         # a pipeline object.
         config = self.pipeline.get_software_configuration('nu-qc')
 
+        if not hasattr(self, 'files_regex'):
+            self.files_regex = 'SPP'
```
AmandaBirmingham, line 482: What does this SPP regex match?  Some sort of internal files produced by a partial SPP run?

src/qp_klp/StandardMetagenomicWorkflow.py
```diff
@@ -146,15 +194,19 @@ def __init__(self, **kwargs):
                 ValueError(f'The run_prefix {rp} from {k} has {_d.shape[0]} '
                            'matches with files')
 
+            # making sure that the index are unique for each row; note that
+            # this is required for make_sample_sheet but ignored downstream
             if 'index' not in vals:
-                vals['index'] = ''
+                vals['index'] = 'G'*i
```
AmandaBirmingham, line 200: What do the "G" and "A" signify?

AmandaBirmingham (changes requested):
A couple of typos; other than that, just some requests for comment and questions

PR #134, opened 2025-07-07
remove sequencer types info, instead get from kl-metapool

AmandaBirmingham:
WILL BREAK UNTIL kl-metapool PR IS ACCEPTED & INSTALLED

This is my suggested change to remove the sequencer type definitions, and the code to get machine prefix from instrument id, from this repo and localize it entirely in kl-metapool.  I have not been able to run tests for this in my dev environment so it should be looked at very carefully :)

### DIFF
--- a/src/sequence_processing_pipeline/Pipeline.py
+++ b/src/sequence_processing_pipeline/Pipeline.py
@@ -6,7 +6,7 @@ from os.path import join, exists, isdir, basename
 from metapool import (load_sample_sheet, AmpliconSampleSheet, is_blank,
                       parse_project_name, SAMPLE_NAME_KEY, QIITA_ID_KEY,
                       PROJECT_SHORT_NAME_KEY, PROJECT_FULL_NAME_KEY,
-                      CONTAINS_REPLICATES_KEY)
+                      CONTAINS_REPLICATES_KEY, get_model_by_instrument_id)
 from metapool.plate import ErrorMessage, WarningMessage
 from sequence_processing_pipeline.Job import Job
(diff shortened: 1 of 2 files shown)


AmandaBirmingham:
I see that three of the tests are failing because of the change to how sample sheets are loaded (which were not part of the sequencer removal, but rather part of immediately-previous kl-metapool work that fixed the inability to load AbsQuantv11 sample sheets).  

The failing tests and the errors are shown below.  I think that the errors that are currently being given ("X is not an unrecognized Assay/SheetType") is better than the previous error, since it gives some guidance as to what the actual problem is.  If you agree, we should change the tests to look for the new, more specific errors.  @antgonza What do you think?

FAIL: test_assay_value (tests.test_NuQCJob.TestNuQCJob)
----------------------------------------------------------------------
AssertionError: "bad-sample-sheet-metagenomics.csv' does not appear to be a valid sample-sheet." does not match "'Metagenomics' is an unrecognized Assay type"

======================================================================
FAIL: test_nuqcjob_creation (tests.test_NuQCJob.TestNuQCJob)
----------------------------------------------------------------------
AssertionError: "tmp-sample-sheet.csv' does not appear to be a valid sample-sheet." does not match "'NotMetagenomic' is an unrecognized Assay type"

======================================================================
FAIL: test_invalid_sample_sheets (tests.test_WorkflowFactory.WorkflowFactoryTests)
----------------------------------------------------------------------
AssertionError: "'tests/data/sample-sheets/metagenomic/illumina/bad_sheet1.csv' does not appear to be a valid sample-sheet." does not match "'not_a_metag' is an unrecognized SheetType"

PR #138, opened 2025-07-10
add subsample_reads

antgonza:


src/qp_klp/Protocol.py
```diff
@@ -22,6 +25,48 @@ class Protocol():
...
+                    nbn = join(dn, bn.replace('fastq.gz', 'subsampled.gz'))
+                    cmd = f'mv {f} {nbn}'
+                    _, se, rv = system_call(cmd)
+                    if rv != 0 or se:
+                        raise ValueError(f'Error during mv: {cmd}. {se}')
+                    cmd = (f'seqtk sample -s 42 {nbn} {self.MAX_READS} '
+                           f'| gzip > {f}')
+                    _, se, rv = system_call(cmd)
+                    if rv != 0 or se:
+                        raise ValueError(f'Error during mv: {cmd}. {se}')
```
AmandaBirmingham, line 66: This error message seems misleading since it (still) says "Error during mv"; I think it should be changed to say it is reporting on errors that occur during `seqtk`.

src/qp_klp/Protocol.py
```diff
@@ -22,6 +25,48 @@ class Protocol():
...
+                'please let an admin know.')
+        df = df[df[read_col] > self.MAX_READS]
+        if df.shape[0]:
+            for _, row in df.iterrows():
+                sn = row[index_col]
+                files = glob(f'{self.raw_fastq_files_path}/*/{sn}*.fastq.gz')
+                for f in files:
+                    dn = dirname(f)
+                    bn = basename(f)
+                    nbn = join(dn, bn.replace('fastq.gz', 'subsampled.gz'))
```
AmandaBirmingham, line 57: Is it going to cause any problems later that the subsampled reads file doesn't end with 'fastq.gz'? I'm not familiar with the workflow but I know I've seen regexes around that expect this suffix ...

src/qp_klp/Assays.py
```diff
@@ -166,6 +167,7 @@ def execute_pipeline(self):
             self.convert_raw_to_fastq()
             self.integrate_results()
             self.generate_sequence_counts()
+            self.subsample_reads()
```
AmandaBirmingham, line 170: Does this mean that we will now ALWAYS subsample?

AmandaBirmingham (changes requested):
One request for change to an error message, a couple of questions

src/qp_klp/Assays.py
antgonza: Yes, but really we will always check if subsample is needed and only run it when necessary.

AmandaBirmingham: Well, we will always subsample every fastq that has more than the max number of reads, right?

PR #139, opened 2025-07-14
QC_Analysis_TellReadJob.html

antgonza:


tests/test_WorkflowFactory.py
```diff
@@ -222,8 +222,13 @@ def test_metagenomic_workflow_creation(self):
...
         # uses 78901 as the job_id so the db complains about the format
-        with self.assertRaisesRegex(RuntimeError, 'invalid input '
-                                    'syntax for type uuid: "78901"'):
+        # Note 07.17.25: for whatever reason the reponse for this test
+        #                changed and is no longer 'invalid input '
+        #                'syntax for type uuid: "78901"; leaving this
+        #                note for future reference
+        with self.assertRaisesRegex(
+                RuntimeError, "Request 'post https://localhost:21174/"
+                "qiita_db/prep_template/' did not succeed"):
```
AmandaBirmingham, line 231: Antonio, I see that the code that is actually throwing this new error hasn't changed in a while, so it is hard to figure out why it would suddenly start erroring.  However, it looks like the code almost directly above it at line 618, that sets the prep file to be the `original-prep.csv`, is a quite recent addition.  Is there any chance that the original-prep.csv, whatever it is, is missing out on some kind of name scrubbing that the previously-used files in `self.prep_file_paths` went through?

I am asking because the new error we are getting is (obviously) a correct one--there are sample names in the test input file that DO contain invalid characters.  Thus, what I am really puzzled about is why we WEREN'T getting this error before ... where did the previous `invalid input syntax for type uuid` error get thrown when this test was working as expected?  Was it upstream or downstream of the error we are seeing now?

antgonza: I'm not sure what the problem could be. I have tried multiple times in my local installation and it works fine. I have a couple of ideas that I'm going to try.

AmandaBirmingham (approved):
WOW, I would definitely never have guessed that cause.  Serious props for tracking that down!!

PR #143, opened 2025-09-01
PacBio

antgonza:


tests/test_Pipeline.py
```diff
@@ -207,43 +207,6 @@ def test_get_orig_names_from_sheet_with_replicates(self):
 
         self.assertEqual(set(obs), exp)
 
-    def test_required_file_checks(self):
```
antgonza, line 210: These test were (1) specific to Illumina so they should not be part of Pipeline, (2) duplication of what normally happens in any code: check if a files exists and raise an error; thus, in my opinion, unnecessary.

antgonza:
Thank you for all the help @AmandaBirmingham

PR #149, opened 2025-11-18
Zed styling

antgonza:


antgonza:
No actual changes, just default Zed changes and flake8 to ruff.

